PR #78, opened 2019-04-23
Implemented leaky tanh with test

koenhelwegen:


larq/activations_test.py
```diff
@@ -20,3 +20,18 @@ def test_hard_tanh():
     f = tf.keras.backend.function([x], [lq.activations.hard_tanh(x)])
     result = f([real_values])[0]
     np.testing.assert_allclose(result, np.clip(real_values, -1, 1))
+
+
+def test_leaky_tanh():
```
lgeiger: Could you add it to `test_serialization` as well?

larq/activations_test.py
```diff
@@ -23,12 +23,14 @@ def test_hard_tanh():
...
 def test_leaky_tanh():
+    @np.vectorize
     def leaky_tanh(x):
-        return (
-            np.clip(x, -1, 1)
-            + (np.minimum(x, -1) + 1) * 0.2
-            + (np.maximum(x, 1) - 1) * 0.2
-        )
+        if x <= -1:
+            return -1 + .2*(x+1)
```
lgeiger: Could you run `black` over this?

PR #81, opened 2019-04-23
Plot output and gradients of activations and quantizers in docs

lgeiger:
This adds simple matplotlib plots to our documentation, which will also add an additional sanity check for the math:
Bildschirmfoto 2019-04-23 um 21 06 53


koenhelwegen:
Awesome!! This is really cool. Small request: would it be possible to add titles? E.g. "Forward pass (activation)" and "Backward pass (gradient)" or something like that? The dy/dx is a bit too subtle I think.

lgeiger:
Added titles and modified the font to match the docs theme. It's not perfect should look reasonable:
Bildschirmfoto 2019-04-24 um 10 17 16

PR #85, opened 2019-05-01
Test QuantizationLogger with TF 2 nightly

lgeiger:
This previously failed due to a bug in tensorflow 2.0.0-alpha0

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,13 +12,13 @@ jobs:
       matrix:
         Python36:
           python.version: "3.6"
-          tensorflow.version: "1.13.1"
+          tensorflow: "tensorflow==1.13.1"
         Python37:
           python.version: "3.7"
-          tensorflow.version: "1.13.1"
+          tensorflow: "tensorflow==1.13.1"
         Python37TF2:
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -12,13 +12,13 @@ jobs:
...
-          tensorflow.version: "1.13.1"
+          tensorflow: "tensorflow==1.13.1"
         Python37:
           python.version: "3.7"
-          tensorflow.version: "1.13.1"
+          tensorflow: "tensorflow==1.13.1"
         Python37TF2:
           python.version: "3.7"
-          tensorflow.version: "2.0.0-alpha0"
+          tensorflow: "tf-nightly-2.0-preview"
```
koenhelwegen: What exactly is affected by this? I would prefer not having to use the nightly build.

lgeiger: This is our CI on Azure Pipelines that executes the tests for TF 2.

`2.0.0-alpha0` has a bug that prevents the `QuantizationLogger` from working properly. This is fixed in the latest nightly. I think in any case it makes sense to test against the nightly build, so we can see when things break

PR #86, opened 2019-05-08
Ternary

jamescook106:
Ternarization. Currently still need to do unit tests before can be merged.


larq/quantizers.py
```diff
@@ -130,6 +130,94 @@ def approx_sign(x):
...
+    """
+
+    def __init__(self, threshold_value=0.1, ternary_weight_networks=False):
+        self.threshold_value = threshold_value
+        self.ternary_weight_networks = ternary_weight_networks
+
+    def __call__(self, x):
+        return self.ste_tern(x)
+
+    def ste_tern(self, x):
```
lgeiger: Can you change this to `__call__` since it has the same signature anyway?

larq/quantizers.py
```diff
@@ -130,6 +130,94 @@ def approx_sign(x):
...
+    ternary_weight_networks: Boolean of whether to use the Ternary Weight Networks threshold calculation. 
+
+    # Returns
+    Ternarized tensor.
+
+    # References
+    - [Ternary Weight Networks](http://arxiv.org/abs/1605.04711)
+    """
+
+    def __init__(self, threshold_value=0.1, ternary_weight_networks=False):
```
lgeiger: I think `ternary_weight_networks` param should be a bit more descriptive, though I can't think of a better name on top of my head.

larq/quantizers.py
```diff
@@ -130,6 +130,94 @@ def approx_sign(x):
...
+        def _ternarize_with_identity_grad(x):
+            def grad(dy):
+                return dy
+
+            return (
+                tf.sign(
+                    tf.add(
+                        tf.sign(tf.add(x, threshold)), tf.sign(tf.add(x, -threshold))
+                    )
+                ),
```
lgeiger: This can be simplified to:
```python
tf.sign(tf.sign(x + threshold) + tf.sign(x - threshold))
```

larq/quantizers.py
```diff
@@ -130,6 +130,94 @@ def approx_sign(x):
...
+                grad,
+            )
+
+        return _ternarize_with_identity_grad(x)
+
+    def threshold_twn(self, x):
+        x_sum = tf.reduce_sum(tf.abs(x))
```
lgeiger: Do we need a `tf.stop_gradient` here?

larq/quantizers.py
```diff
@@ -130,6 +130,94 @@ def approx_sign(x):
...
+        threshold = tf.div(x_sum, tf.cast(tf.size(x), tf.float32))
```
lgeiger: We should cast it to the `dtype` of the input tensor:
```suggestion
        threshold = x_sum / tf.cast(tf.size(x), x.dtype)
```

larq/quantizers.py
```diff
@@ -130,6 +130,94 @@ def approx_sign(x):
...
+        threshold = tf.multiply(0.7, threshold)
```
```suggestion
        threshold = 0.7 * threshold
```

larq/quantizers_test.py
```diff
@@ -51,6 +51,26 @@ def test_binarization(name):
     assert np.all(result[result >= 0] == 1)
 
 
+def test_ternarization():
+    x = tf.keras.backend.placeholder(ndim=2)
+    test_threshold = np.random.uniform(0.01, 0.8, 1)[0]
```
lgeiger:
```suggestion
    test_threshold = np.random.uniform(0.01, 0.8)
```

jamescook106: Updated in latest commit to this

larq/quantizers.py
jamescook106: I am unsure, I briefly looked up the documentation and I don't think we do, but there is a significant chance that I am incorrect

larq/quantizers.py
jamescook106: Implemented in latest commit

larq/quantizers.py
jamescook106: Implemented in latest commit

larq/quantizers.py
jamescook106: Agreed and updated in latest commit

larq/quantizers.py
jamescook106: Updated in latest commit

larq/quantizers.py
jamescook106: I'm happy with this name as it's the original paper name, and I only think they use this method in the original paper

jamescook106:
@lgeiger I have responded to all of your review comments, please look at the azure pipeline when you have time

PR #93, opened 2019-05-10
Display links to source code

### LINKED ISSUE #88: Docs: Add links to source code
This is really handy if people want to understand what's going on behind the scenes or want to implement more advanced stuff

lgeiger:
This will add links to the source code on hover:
Bildschirmfoto 2019-05-10 um 15 01 07

~It's not perfect. For functions it links to the line of the decorator (one line above the actual function, also `inspect.unwrap` didn't help), but it's still a large improvement.~ fixed in 0ec98ba8f5fb5a6240df6e23575c2c5249f6267c
Fixes #88

### DIFF
--- a/docs/custom.css
+++ b/docs/custom.css
@@ -1,0 +1,3 @@
+.code-link::before {
+  content: "</>";
+}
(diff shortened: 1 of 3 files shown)


docs/custom.css
```diff
@@ -0,0 +1,3 @@
+.code-link::before {
+  content: "</>";
```
lgeiger: This is a bit hacky.

We add the text via the `::before` element in order to prevent mkdocs from adding it to the table of contents as well.

PR #96, opened 2019-05-28
Lazily compute quantized_weights

lgeiger:
We are running into strange GPU OOM errors.

@jamescook106 Could you test if this worksaround the TensorFlow issue?

In the future I'd be in favor of removing the `quantized_weights` and `quantized_latent_weights` properties. This PR only makes them lazily computed in order to not break the summary and xavier rescaling.

### DIFF
--- a/larq/layers.py
+++ b/larq/layers.py
@@ -34,14 +34,17 @@ class QuantizerBase(tf.keras.layers.Layer):
                 "may result in starved weights (where the gradient is always zero)."
             )
 
-        self.quantized_weights = []
-        self.quantized_latent_weights = []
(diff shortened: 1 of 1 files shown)


jamescook106:
Testing as we speak. Note the hasattr(l, "quantized_weights") syntax still works

jamescook106:
This does not fix the bug

jamescook106:
Screenshot from 2019-05-29 10-31-20

lgeiger:
It looks like your memory error is unrelated. I still think this PR is valuable. I will go ahead and merge it.

PR #99, opened 2019-06-07
Prevent QuantizedLogger from adding nodes to the graph during training

lgeiger:
#96 Prevented unnecessary nodes from being created during layer build time if the logger isn't used. Unfortunately this has the effect that `QuantizedLogger` constantly adds new nodes to the graph during logging since the `quantized_weights` property is now layzily computed and doesn't reuse the already existing ops in the graph (I don't know if there is a way to force this reuse, `tf.variable_scope(..., reuse=True)` only works for variables).

This PR fixes the problem by only calling `layer.quantized_weights` once during callback initialization.

In the future I'd really like to do this in a proper way. Keras layers have an [`add_metric` method](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/engine/base_layer_test.py#L982-L992) that would allow us to add metrics like this. Unfortunately the metrics we are interested in are stateful since they compare pre and post-update states so we need to properly benchmark this to see if a solution like this would come with performance penalties sinc metrics are accumulated at each update step.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -18,7 +18,7 @@ jobs:
           tensorflow: "tensorflow==1.13.1"
         Python37TF2:
           python.version: "3.7"
-          tensorflow: "tf-nightly-2.0-preview"
+          tensorflow: "tensorflow==2.0.0b0"
           coverage: "true"
 
     steps:
--- a/larq/callbacks.py
+++ b/larq/callbacks.py
@@ -27,21 +27,27 @@ class QuantizationLogger(tf.keras.callbacks.Callback):
     """
 
     def __init__(self, update_freq="epoch"):
+        super().__init__()
         self.batch_previous_weights = {}
         self.epoch_previous_weights = {}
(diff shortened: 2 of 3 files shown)


azure-pipelines.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           tensorflow: "tensorflow==1.13.1"
         Python37TF2:
           python.version: "3.7"
-          tensorflow: "tf-nightly-2.0-preview"
+          tensorflow: "tf-nightly-2.0-preview==2.0.0.dev20190603"
```
lgeiger: Looks like the newest nightly has some build problems. I pinned it for now, so our builds are passing again.

lgeiger: Tensorflow just released their [beta release](https://github.com/tensorflow/tensorflow/releases/tag/v2.0.0-beta0). I'll pin it to that.

arashb:
I am not sure if the previous code was adding really extra nodes to Graph but this code is anyway more efficient than before so lGTM.

PR #100, opened 2019-06-10
Add layer metrics to replace quant logger callback

lgeiger:
This is an early prototype to explore the Keras `layers.add_metrics` API for our stateful metrics as a replacements for the `QuantizedLogger`. I think this approach is a lot more flexible and elegant than using a callback.

Before I'll clean this PR up, we should check if this approach has a big performance impact since it needs to keep track of the previous quantized weights.

### DIFF
--- a/larq/callbacks_test.py
+++ b/larq/callbacks_test.py
@@ -7,7 +7,6 @@ from larq.callbacks import HyperparameterScheduler
 
 from tensorflow import keras
 from tensorflow.python.keras import testing_utils
-from tensorflow.python.keras import keras_parameterized
 
 
 class LogHistory(tf.keras.callbacks.Callback):
@@ -26,8 +25,7 @@ class LogHistory(tf.keras.callbacks.Callback):
         self._store_logs(self.epochs, epoch, logs)
 
 
-@keras_parameterized.run_all_keras_modes
-class LayersTest(keras_parameterized.TestCase):
+class TestQuantizationLogger:
     def test_quantization_logger(self):
         model = tf.keras.models.Sequential(
             [
--- a/larq/layers.py
+++ b/larq/layers.py
@@ -21,9 +21,17 @@ class QuantizerBase(tf.keras.layers.Layer):
     equivalent to `Layer`.
     """
 
-    def __init__(self, *args, input_quantizer=None, kernel_quantizer=None, **kwargs):
+    def __init__(
+        self,
+        *args,
+        input_quantizer=None,
+        kernel_quantizer=None,
+        metrics=["mean_changed_values"],
+        **kwargs,
+    ):
         super().__init__(*args, **kwargs)
 
+        self._custom_metrics = metrics
         self.input_quantizer = quantizers.get(input_quantizer)
         self.kernel_quantizer = quantizers.get(kernel_quantizer)
 
@@ -35,8 +43,8 @@ class QuantizerBase(tf.keras.layers.Layer):
 
     def build(self, input_shape):
         super().build(input_shape)
-        # TODO make configurable
-        self.mean_changed_values = metrics.MeanChangedValues(self.kernel.shape)
+        if "mean_changed_values" in self._custom_metrics:
+            self.mean_changed_values = metrics.MeanChangedValues(self.kernel.shape)
 
     @property
     def quantized_weights(self):
@@ -57,7 +65,8 @@ class QuantizerBase(tf.keras.layers.Layer):
         if self.kernel_quantizer:
             full_precision_kernel = self.kernel
             quantized_kernel = self.kernel_quantizer(self.kernel)
-            self.add_metric(self.mean_changed_values(quantized_kernel))
+            if "mean_changed_values" in self._custom_metrics:
+                self.add_metric(self.mean_changed_values(quantized_kernel))
             self.kernel = quantized_kernel
 
         output = super().call(inputs)
--- a/larq/layers.py
+++ b/larq/layers.py
@@ -7,7 +7,6 @@ is equivalent to a full precision layer.
 """
 
 import logging
-from distutils.version import LooseVersion
 import tensorflow as tf
 from larq import quantizers, utils, metrics
 
@@ -17,9 +16,7 @@ log = logging.getLogger(__name__)
 def _supports_metrics():
     # TensorFlow 1.13 does not support adding an aggregated metric tensor in
     # `tf.keras.layers.Layer.call` in eager execution.
-    return not (
-        LooseVersion(tf.__version__) < LooseVersion("1.14.0") and tf.executing_eagerly()
-    )
+    return utils.tf_1_14_or_newer() or not tf.executing_eagerly()
 
 
 class QuantizerBase(tf.keras.layers.Layer):
--- a/larq/layers.py
+++ b/larq/layers.py
@@ -29,7 +29,7 @@ class QuantizerBase(tf.keras.layers.Layer):
 
     def __init__(self, *args, input_quantizer=None, kernel_quantizer=None, **kwargs):
         # This is currently undocumented until we have explored better options
-        self._custom_metrics = kwargs.pop("metrics", ["mean_changed_values"])
+        self._custom_metrics = kwargs.pop("metrics", ["flip_ratio"])
 
         self.input_quantizer = quantizers.get(input_quantizer)
         self.kernel_quantizer = quantizers.get(kernel_quantizer)
@@ -45,11 +45,11 @@ class QuantizerBase(tf.keras.layers.Layer):
         super().build(input_shape)
         if (
             self.kernel_quantizer
-            and "mean_changed_values" in self._custom_metrics
+            and "flip_ratio" in self._custom_metrics
             and _supports_metrics()
         ):
-            self.mean_changed_values = metrics.MeanChangedValues(
-                values_shape=self.kernel.shape, name=f"{self.name}/mean_changed_values"
+            self.flip_ratio = metrics.FlipRatio(
+                values_shape=self.kernel.shape, name=f"{self.name}/flip_ratio"
             )
 
     @property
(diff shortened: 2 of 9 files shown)


lgeiger:
I tested the the performance impact of this approach for Bi-Real Net and BinaryNet and found that it is almost negligible with batch size 256:
The runtime increases approximately by 1% (which is inside the margin of my ad-hoc performance testing) and memory usage only increases by less than 0.01-0.1%.

I tested maximum memory usage using:
```python
from tensorflow.contrib.memory_stats.python.ops.memory_stats_ops import MaxBytesInUse

class MemoryPrinter(tf.keras.callbacks.Callback):
    def on_train_begin(self, logs={}):
        with tf.device("/device:GPU:0"):
            self.bytes_in_use = MaxBytesInUse()

    def on_epoch_end(self, epoch, logs={}):
        bytes = tf.keras.backend.get_value(self.bytes_in_use)
        print(f"Epoch {epoch}, bytes {bytes}")
```
and speed just by looking at the execution time over a few epochs, which isn't really accurate but should give a good indication of how this approach performs.

I think this could allow us to clean up a few internal implementation details and add support for properly averaged metrics that are computed over multiple batches.

@koenhelwegen @jamescook106 @arashb Let's discuss tomorrow if this is a good approach to take and would be easier to use than the current `QuantizationLogger` approach.

lgeiger:
We decided to move forward with this approach and 🔥 the `QuantizationLogger`. Here are the things that are left to do:
- [x] Add layer names to logged metrics
- [ ] Add support for separable layers
- [ ] Fix current layer unit tests
- [ ] Add unit tests for the new metrics
- [ ] Think of a better name for our first metric. I opted for `mean_changed_values` though I am not sure if it is the most descriptive. @jamescook106 @koenhelwegen what do you think?
- [ ] Explore ways to configure layer metrics from `model.compile` or `model.fit` without needing to mess with the layer keyword arguments (Might be done in a future PR).

larq/testing_utils.py
```diff
@@ -27,3 +32,172 @@ def get_small_bnn_model(input_dim, num_hidden, output_dim):
...
+    model_config = model.get_config()
+    recovered_model = tf.keras.models.Model.from_config(model_config)
+    if model.weights:
+        weights = model.get_weights()
+        recovered_model.set_weights(weights)
+        output = recovered_model.predict(input_data)
+        np.testing.assert_allclose(output, actual_output, rtol=2e-3)
+
+    # Recreate layer to prevent layer metrics from being configured multiple times.
+    layer = layer_cls(**kwargs)
```
lgeiger: This is the line, why we need to fork this function. In general I think it is not too bad to maintain our own layer tests to have full control.

larq/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+
+    def get_config(self):
+        return {
+            **super().get_config(),
+            "values_shape": self.values_shape,
+            "values_dtype": self.values_dtype.name,
+        }
+
+    def add_weight(
```
lgeiger: All of this code is really annoying, but the alternative would be not to cast the previous weights to a smaller dtype which might or might not increase memory consumption.

larq/optimizers_test.py
```diff
@@ -40,7 +39,8 @@ def _test_optimizer(optimizer, target=0.75, test_kernels_are_binary=True):
     if test_kernels_are_binary:
         for layer in model.layers:
             if "quant" in layer.name:
-                assert np.all(np.isin(layer.get_weights(), [-1, 1]))
+                for weight in layer.trainable_weights:
+                    assert np.all(np.isin(tf.keras.backend.get_value(weight), [-1, 1]))
```
lgeiger: Metrics are added as non-trainable variables to a layer which aren't binary.

larq/callbacks_test.py
```diff
@@ -20,29 +19,27 @@ def on_epoch_end(self, epoch, logs={}):
...
+                pointwise_quantizer="ste_sign",
+                input_shape=(3, 3),
+            ),
+            tf.keras.layers.Flatten(),
+            lq.layers.QuantDense(2, kernel_quantizer="ste_sign"),
+        ]
+    )
+    model.compile(loss="mse", optimizer="sgd")
+
+    logger = lq.callbacks.QuantizationLogger(update_freq=5)
```
lgeiger: Let's not 🔥 the `QuantizationLogger` in this PR to keep the diff manageable.

larq/metrics.py
```diff
@@ -0,0 +1,99 @@
+import tensorflow as tf
+from larq import utils
+import numpy as np
+
+
+class MeanChangedValues(tf.keras.metrics.Mean):
```
lgeiger: @koenhelwegen @jamescook106 What do you think about the name, I am not sure if it is the best we can come up with.

lgeiger: I didn't expose the metric in the public API and docs for now, so we can change it if we need to. I would like to make this metric lazy and not require users to pass in a static shape during construction, but this needs more thinking to prevent unnecessary nodes from being added to the graph.

larq/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+        values = tf.cast(values, self.values_dtype)
+        if not self._built:
+            self._built = True
+            return self._previous_values.assign(values)
+        changed_values = tf.math.count_nonzero(tf.equal(self._previous_values, values))
+        metric_update_op = super(MeanChangedValues, self).update_state(
+            1 - (tf.cast(changed_values, self.dtype) / self._size)
+        )
+        with tf.control_dependencies([metric_update_op]):
+            return self._previous_values.assign(values)
```
koenhelwegen: Maybe I'm missing something, but I have strong doubts about this. This function returns an op, but it assumes the op is executed only once and `update_state` is called again every time. In eager execution I guess this is fine? The tests follow this pattern, yielding the proper results (but adding nodes to the graph for every `update_state` call in graph mode). In general I would expect this function to create the necessary op which I can then simply execute, i.e. call it only once. I'm not sure how Keras interacts with this.

Perhaps we can initialize `_previous_values` with zeros and ignore the first ratio by using something like `total += ratio*(count > 0)`? This would require implementing the streaming logic ourselves.

koenhelwegen: See also [`keras.metrics.MeanTensor`](https://github.com/tensorflow/tensorflow/blob/3699977134badccb7032fa6921d70e01ba8fdf7d/tensorflow/python/keras/metrics.py#L2349) which also uses a `_built` pattern but avoids returning different ops

larq/layers.py
```diff
@@ -52,7 +70,10 @@ def call(self, inputs):
             inputs = self.input_quantizer(inputs)
         if self.kernel_quantizer:
             full_precision_kernel = self.kernel
-            self.kernel = self.kernel_quantizer(self.kernel)
+            quantized_kernel = self.kernel_quantizer(self.kernel)
+            if hasattr(self, "mean_changed_values"):
+                self.add_metric(self.mean_changed_values(quantized_kernel))
```
koenhelwegen: Could you perhaps add a brief comment on how this works?

larq/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+        self,
+        values_shape=(),
+        values_dtype="int8",
+        name="mean_changed_values",
+        dtype=None,
+    ):
+        super().__init__(name=name, dtype=dtype)
+        self.values_dtype = tf.as_dtype(values_dtype)
+        self.values_shape = tf.TensorShape(values_shape).as_list()
+        self._previous_values = self.add_weight(
```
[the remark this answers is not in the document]
lgeiger: I assumed that namescopes will be handled automatically. Which is not the case here. Good catch :+1:

larq/metrics.py
lgeiger: I like `FlipRatio`. Other options would be `ChangeRatio` or `AlteringRatio`.

larq/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+class MeanChangedValues(tf.keras.metrics.Mean):
+    """Computes the mean ration of changed values in a given tensor.
+
+    !!! example
+        ```python
+        m = metrics.MeanChangedValues()
+        m.update_state(1)
+        m.update_state(2)
+        m.update_state(2)
+        print('Final result: ', m.result().numpy())  # Final result: 0.5
```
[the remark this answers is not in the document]
lgeiger: :+1: I'll add the example I use in the unit tests

larq/layers.py
lgeiger: It adds the metric to a layer the same way an accuracy metric is added to a model in when passing it to `keras.models.Model.compile` (If I recall correctly, `keras.models.Model` also inherits from `keras.layers.Layer`):

https://github.com/tensorflow/tensorflow/blob/0bbeb15d2247f2fae98cbec4e5a93c285ed5ab0c/tensorflow/python/keras/engine/base_layer.py#L1026-L1042

larq/metrics.py
lgeiger: I had a close look at `keras.metrics.MeanTensor`, because I like that it doesn't require users to initialize the metric with a known shape.
Actually I tried the same approach with lazily calling `add_weight` in `_build` on the first execution of the metric, but that resulted in unnecessary nodes in the graph during eager execution when adding it to a layer. That's a thing I'd like to debug in the future. As for `update_state`, `keras.metrics.MeanTensor` actually uses the same pattern as we (and all other metrics) do here and returns an assign op:
https://github.com/tensorflow/tensorflow/blob/3699977134badccb7032fa6921d70e01ba8fdf7d/tensorflow/python/keras/metrics.py#L2403-L2444

In graph mode, Keras actually sets up a internal conceptual graph and will only add the update ops once. As for eager mode, the metrics is evaluated lazily similar to the pattern you see in the unit tests.

I am confident, that our layer unit tests would fail if we add the metric multiple times.

koenhelwegen: My problem is that unlike any other Keras metric we return two different assign ops, and it is essential they are used both. If keras only creates the ops once (which is indeed what I would expect), it gets the op defined under `if not self._built` and the streaming metric is never updated.

larq/metrics.py
lgeiger: Changed it to `FlipRatio`

larq/metrics.py
lgeiger: You are correct, good catch 👍 

I changed the implementation in 096a6e0

lgeiger:
@koenhelwegen This is now ready for another critical look.

Issue #101, opened 2019-06-12
Add support for XavierLarningRateScaling in TF 1.14+


arashb:
Optimizers tests are failing with TF 1.14.0rc0 and TF 1.14.0rc1. 

Error output: 

```
self = <larq.optimizers.XavierLearningRateScaling object at 0x133f54e80>, name = 'optimizer'
                                                                                                                                                             
    def __getattribute__(self, name):                                                                                                                        
      """Overridden to support hyperparameter access."""
      try:                            
>       return super(OptimizerV2, self).__getattribute__(name)
E       AttributeError: 'XavierLearningRateScaling' object has no attribute 'optimizer'
                                      
../venv-bop/lib/python3.7/site-packages/tensorflow/python/keras/optimizer_v2/optimizer_v2.py:532: AttributeError
```

arashb:
@lgeiger can you add a "XavierLarningRateScaling" class to "optimizers_v2.py" with v2 API?

lgeiger:
> @lgeiger can you add a "XavierLarningRateScaling" class to "optimizers_v2.py" with v2 API?

I had a quick shot at this when I added `XavierLarningRateScaling` but it didn't work at first try. At this point I am unsure if `XavierLarningRateScaling` is really that useful in general and is mostly here because it is frequently used in the earlier literature of BNNs. I think it would make sense to give some general way of defining separate learning rates for quantized and non quantized parameters, but for now this doesn't have high priority for me.

arashb:
Beside implementing the new optimizer API, do you see any other difficulties to get this working with TF 1.14+? we have already updated the API  for Bop optimizer and it shouldn't be a huge hassle, IMO.

lgeiger:
> Beside implementing the new optimizer API, do you see any other difficulties to get this working with TF 1.14+? we have already updated the API for Bop optimizer and it shouldn't be a huge hassle, IMO.

I also encountered errors due to [these lines]( https://github.com/plumerai/larq/blob/master/larq/optimizers.py#L84-L85), but that might be because I was trying to support both v1 and v2 in one implementation. I can push my WIP branch from some time ago and we can take a look and see if there is a quick fix 👍

lgeiger:
[...]

PR #102, opened 2019-06-13
Skipping the tests for Xavier Learning Rate Scaling optimizer for TFversions >1.13 (#101)

arashb:
The optimizer interface has changed from version 1.13 to +1.14. This patch as a temporary solution, skips the Xavier Learning Rate Scaling tests for incompatible TF versions until we rewrite the Xavier optimizer for new optimizer API.


arashb:
This issue is resolved in Bop PR.

PR #103, opened 2019-06-14
Adding Bop optimizer

arashb:
In this patch two issues are addressed:

1. we concluded we need two separate implementation of optimizers for TF <=v1.13 and TF >=v1.14. Therefore we have two separate files (optimizers_v1|2.py) for each implementation and the decision to import which one is made in optimizers.py file.

2. I added the TF >v1.14 Bop implementation to optimizers_v2.py and a modified but not yet functional version to optimizers_v1.py. In addition, a few unittests for Bop is added to optimizers_test.py.


koenhelwegen:
I added a working version for 1.13. The only thing I'm not sure about is hyperparameter schedules, both for Bop and for the fp_optimizer. Marked the PR as draft until that is resolved.

arashb:
with the latest commit, I added the Bop learning rate scheduler. The tests pass for both TF 1.13 and 1.14. Not tested for TF 2.0 yet.

larq/callbacks.py
```diff
@@ -1,6 +1,9 @@
 import tensorflow as tf
 import numpy as np
 
+from larq.optimizers import Bop
+from tensorflow.python.keras import backend as K
```
lgeiger: In general, we should try not to import from `tensorflow.python` since this is not a public API and might change.

larq/callbacks.py
```diff
@@ -68,3 +71,47 @@ def on_train_begin(self, logs=None):
 
     def on_epoch_end(self, epoch, logs=None):
         self._maybe_log_and_store(self.epoch_previous_weights, logs)
+
+
+class BopLearningRateScheduler(tf.keras.callbacks.Callback):
```
lgeiger: I don't think we should add a new callback just dedicated for Bop. I think it would be could to find ways of handling this usecase inside the Bop class.

larq/callbacks_test.py
```diff
@@ -1,7 +1,14 @@
 import numpy as np
 import tensorflow as tf
 import larq as lq
+
+
+from larq import testing_utils as lq_testing_utils
+from larq.callbacks import BopLearningRateScheduler
+
+from tensorflow.python import keras
```
lgeiger: I think using `tf.keras` directly or `from tensorflow import keras` would be better since `tensorflow.pyhon` is a private API.

larq/callbacks_test.py
```diff
@@ -46,3 +53,42 @@ def test_quantization_logger(self):
...
+            batch_size=16,
+            validation_data=(x_test, y_test),
+            callbacks=cbks,
+            verbose=0,
+        )
+        np.testing.assert_almost_equal(
+            float(keras.backend.get_value(model.optimizer.fp_optimizer.lr)),
+            scheduler(num_epochs - 1),
+            decimal=8,
+        )
```
lgeiger: Awesome! Can we also assert that all weights in the model are still binary? `model.get_weights()` will give you the numpy array of all weights.

larq/optimizers_v1.py
```diff
@@ -0,0 +1,180 @@
...
+import numpy as np
+import larq as lq
+
+from larq import utils
+from copy import deepcopy
+
+from tensorflow.python.keras.optimizers import *
+from tensorflow.python.framework import ops
+from tensorflow.python.ops import random_ops
+from tensorflow.python.keras import backend as K
```
lgeiger: We should try not to import from `tensorflow.python` as best as we can since those imports might changes (and `*` might give us very weird behaviour)

larq/optimizers_v1.py
```diff
@@ -0,0 +1,180 @@
...
+    def __init__(self, fp_optimizer, threshold=1e-5, gamma=1e-2, name="Bop", **kwargs):
+        super().__init__(**kwargs)
+
+        if not isinstance(fp_optimizer, tf.keras.optimizers.Optimizer):
+            raise TypeError(
+                f"Expected tf.keras.optimizers.Optimizer, received {type(fp_optimizer)}."
+            )
+
+        with K.name_scope(self.__class__.__name__):
+
```
lgeiger:
```suggestion
```

larq/optimizers_v1.py
```diff
@@ -0,0 +1,180 @@
...
+            self.fp_optimizer = fp_optimizer
+            self.threshold = threshold
+            self.gamma = gamma
+            self.iterations = K.variable(0, dtype="int64", name="iterations")
```
lgeiger: I don't think `self.iterations` is used anywhere. So we can remove it:
```suggestion
```

larq/optimizers_v1.py
```diff
@@ -0,0 +1,180 @@
...
+                self.updates.append(
+                    state_ops.assign(
+                        p, lq.quantizers.sign(-p * tf.sign(p * m_t - self.threshold))
+                    )
+                )
+
+            else:
+                fp_params.append(p)
+
+        self.updates.append(state_ops.assign(self.iterations, self.iterations + 1))
```
lgeiger: See above
```suggestion
```

larq/optimizers_v2.py
```diff
@@ -0,0 +1,83 @@
...
+        self._set_hyper("gamma", gamma)
+
+    def _create_slots(self, var_list):
+        for var in var_list:
+            if self.is_binary(var):
+                self.add_slot(var, "m")
+
+    def apply_gradients(self, grads_and_vars, name=None):
+        bin_grads_and_vars = [(g, v) for g, v in grads_and_vars if self.is_binary(v)]
+        fp_grads_and_vars = [(g, v) for g, v in grads_and_vars if not self.is_binary(v)]
```
lgeiger: I think we should also allow people to manually pass in the binarized parameters, similar to how they do it [here](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/optimizers/weight_decay_optimizers.py#L131) so people can use this optimizer with a custom training loop and custom layers. I don't think we should add it in this PR though, just some thing to keep in mind for future improvements.

larq/optimizers_v1.py
```diff
@@ -0,0 +1,180 @@
...
+        grads = self.get_gradients(loss, params)
+        self.updates = []
+
+        ms = [K.zeros(K.int_shape(p), dtype=K.dtype(p)) for p in params]
+
+        fp_params = []
+
+        for p, g, m in zip(params, grads, ms):
+            if "/kernel" in p.name and "quant_" in p.name:
+                print("Applying Bop to %s." % p.name)
```
lgeiger: I think we should remove this debug output. It can be quite noise with large models.
```suggestion
```

larq/optimizers_v1.py
```diff
@@ -0,0 +1,180 @@
...
+            "fp_optimizer": {
+                "class_name": self.fp_optimizer.__class__.__name__,
+                "config": fp_optimizer_config,
+            },
+        }
+        return {**super().get_config(), **config}
+
+    @classmethod
+    def from_config(cls, config, custom_objects=None):
+        new_config = deepcopy(config)
```
lgeiger: Do we really need a `deepcopy` here, or is `config = config.copy()` enough?

larq/optimizers_v2.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def _get_decayed_hyper(self, name, var_dtype):
+        hyper = self._get_hyper(name, var_dtype)
+        if isinstance(hyper, tf.keras.optimizers.schedules.LearningRateSchedule):
+            local_step = tf.cast(self.iterations, var_dtype)
+            hyper = tf.cast(hyper(local_step), var_dtype)
+        return hyper
+
+    def _resource_apply_dense(self, grad, var):
+        print(f"Applying Bop to {var}")
```
lgeiger: See above
```suggestion
```

larq/optimizers_v2.py
```diff
@@ -0,0 +1,83 @@
...
+            "fp_optimizer": {
+                "class_name": fp_optimizer_config["name"],
+                "config": fp_optimizer_config,
+            },
+        }
+        return {**super().get_config(), **config}
+
+    @classmethod
+    def from_config(cls, config, custom_objects=None):
+        new_config = deepcopy(config)
```
lgeiger: Do we really need a `deepcopy` here, or is `config = config.copy()` enough?

larq/testing_utils.py
```diff
@@ -0,0 +1,27 @@
+import larq as lq
+
+from tensorflow.python import keras
```
lgeiger:
```suggestion
from tensorflow import keras
```

larq/callbacks.py
arashb: I replaced the K with tf.keras.backend everywhere in the code.

larq/optimizers_v1.py
arashb: need a deepcopy bc fp_optimizer is a nested dictionary.

larq/callbacks_test.py
arashb: ok, changed it.

larq/optimizers_v1.py
arashb: fixed it.

larq/optimizers_v1.py
arashb: I am using iterations in test cases.

larq/callbacks.py
lgeiger: > I replaced the K with tf.keras.backend everywhere in the code.

Looks like it's still imported though

arashb: I haven't push the changes yet.

larq/callbacks_test.py
arashb: @koenhelwegen

koenhelwegen: Yes we should test this, however I would not do it here but in `optimizers_test.py`. Note that this has to happen separately for Bop, as the other optimizers use real-valued latent weights.

larq/optimizers_v2.py
arashb: @koenhelwegen

larq/optimizers_v1.py
lgeiger: > I am using iterations in test cases.

If it is only used in the test case, I think this shouldn't be here unless all Keras optimizers are expected to have an `iterations` attribute.

larq/optimizers_v1.py
lgeiger: > need a deepcopy bc fp_optimizer is a nested dictionary.

True, but we only mutate the top level of the dictionary. Though we can leave it as `deepcopy` to be save :+1:

larq/optimizers_test.py
```diff
@@ -2,49 +2,96 @@
...
+        opt = tf.keras.optimizers.deserialize(config)
+        assert opt.__class__ == ref_opt.__class__
+        assert opt.optimizer.__class__ == ref_opt.optimizer.__class__
+        assert opt.optimizer.get_config() == ref_opt.optimizer.get_config()
+        assert opt.multipliers == ref_opt.multipliers
+
+
+class TestBopOptimizer:
+    def test_bop_accuracy(self):
+        _test_optimizer(lq.optimizers.Bop(fp_optimizer=tf.keras.optimizers.Adam(0.01)))
```
lgeiger: Could you also test that the kernels are still binary?

larq/optimizers_test.py
```diff
@@ -2,49 +2,96 @@
...
+
+    def test_bop_serialization(self):
+        _test_serialization(
```
lgeiger: Is `_test_serialization` used anywhere else? If not, I'd prefer to add the code directly here.

larq/optimizers_v1.py
```diff
@@ -0,0 +1,180 @@
...
+    @classmethod
+    def from_config(cls, config, custom_objects=None):
+        optimizer = tf.keras.optimizers.deserialize(
+            config.pop("optimizer"), custom_objects=custom_objects
+        )
+        return cls(optimizer, config["multipliers"])
+
+
+@utils.register_keras_custom_object
+class Bop(tf.keras.optimizers.Optimizer):
```
lgeiger: Could you add docstrings, so it'll show up in our [API docs](https://plumerai.github.io/larq/api/optimizers/)?

larq/optimizers_v1.py
```diff
@@ -0,0 +1,180 @@
...
+                self.updates.append(
+                    tf.assign(
+                        p, lq.quantizers.sign(-p * tf.sign(p * m_t - self.threshold))
+                    )
+                )
+
+            else:
+                fp_params.append(p)
+
+        self.updates.append(tf.assign(self.iterations, self.iterations + 1))
```
lgeiger: Either remove, or use
```suggestion
        self.updates.append(tf.assign_add(self.iterations, 1))
```

larq/optimizers_v1.py
```diff
@@ -0,0 +1,180 @@
...
+            tf.keras.backend.zeros(
+                tf.keras.backend.int_shape(p), dtype=tf.keras.backend.dtype(p)
+            )
+            for p in params
+        ]
+
+        fp_params = []
+
+        for p, g, m in zip(params, grads, ms):
+            if "/kernel" in p.name and "quant_" in p.name:
```
lgeiger: Either use `self.is_binary(p)` here, or remove `is_binary` because it is not used otherwise.

larq/optimizers_v2.py
```diff
@@ -0,0 +1,82 @@
+import tensorflow as tf
+import numpy as np
+import larq as lq
+
+from larq import utils
+from copy import deepcopy
+
+
+@utils.register_keras_custom_object
+class Bop(tf.keras.optimizers.Optimizer):
```
lgeiger: Could you add docstrings and a link to our paper similar to the xavier learning rate scaling callback?

larq/optimizers_test.py
arashb: this a general function that can be used to test other optimizers too so I would rather keep it.

larq/optimizers_v1.py
```diff
@@ -0,0 +1,195 @@
...
+            self.optimizer.lr = base_lr * self.multipliers[param.name]
+            updates.extend(self.optimizer.get_updates(loss, [param]))
+
+        self.optimizer.lr = base_lr
+        updates.extend(self.optimizer.get_updates(loss, base_lr_params))
+
+        return updates
+
+    def __getattr__(self, name):
+        return getattr(self.optimizer, name)
```
lgeiger: I am not sure if that allows Bop to be used with `tf.keras.callbacks.LearningRateScheduler`. For that to work we'd need to support the `optimizer.lr = 1e-3` syntax. Could you add a test for that and check that it works?

lgeiger: This is fixed by 561eee8

larq/optimizers_test.py
```diff
@@ -90,6 +99,35 @@ class TestBopOptimizer:
     def test_bop_accuracy(self):
         _test_optimizer(lq.optimizers.Bop(fp_optimizer=tf.keras.optimizers.Adam(0.01)))
 
+    def test_bop_lr_scheduler(self):
```
lgeiger: I think we should also add a test for TensorFlow 1.14+  that tests the usage with `tf.keras.optimizer.schedules` similar to what TensorFlow does for their optimizers. But this can be done in a future PR.

arashb:
LGTM

larq/optimizers_test.py
```diff
@@ -2,49 +2,130 @@
...
-    model = tf.keras.models.Sequential([dense])
+    y_train = keras.utils.to_categorical(y_train)
+
+    model = lq_testing_utils.get_small_bnn_model(x_train.shape[1], 20, y_train.shape[1])
+    model.compile(loss="categorical_crossentropy", optimizer=optimizer, metrics=["acc"])
+    history = model.fit(x_train, y_train, epochs=2, batch_size=16, verbose=0)
+
+    for layer in model.layers:
+        if "quant" in layer.name:
+            assert np.all(np.isin(layer.get_weights(), [-1, 1]))
```
koenhelwegen: This fails for optimizers that rely on latent weights (and rely on `kernel_quantizer="some_quantizer"` to obtain a binarized forward pass). In the future we should probably add a flag for whether to run this test.

larq/optimizers_v1.py
```diff
@@ -0,0 +1,200 @@
...
+        fp_params = []
+
+        for p, g, m in zip(params, grads, ms):
+            if self.is_binary(p):
+                m_t = (1 - self.gamma) * m + self.gamma * g
+
+                self.updates.append(tf.assign(m, m_t))
+                self.updates.append(
+                    tf.assign(
+                        p, lq.quantizers.sign(-p * tf.sign(p * m_t - self.threshold))
```
koenhelwegen: @lgeiger I feel the `lq.quantizers.sign` here is a bit confusing, since our quantizers usually modify the backpropagation. How about moving our `sign` to `lq.math.sign` (if we expect to implement more low-precision specific math) or aliasing it to `lq.sign`?

lgeiger: I agree 👍 

I think we should move it to `lq.math.sign`. This is then consistent with TensorFlow which will move a lot of their top level math functions to `tf.math` as well (see [here](https://github.com/tensorflow/community/blob/master/rfcs/20180827-api-names.md#appendix-1-additional-endpoints)).

In the future we can add a heaviside step function to binarize with `{0, 1}` and similar things.

@koenhelwegen Can you make a PR to move this to `lq.math`?

PR #110, opened 2019-06-25
Move sign to lq.math

### LINKED ISSUE #108: Move lq.quantizers.sign to lq.math.sign
As discussed in https://github.com/plumerai/larq/pull/103#discussion_r297192840, we should move `lq.quantizers.sign` to `lq.math.sign` and properly document that:
https://github.com/plumerai/larq/blob/f713c6ba741dc1c0267a3c0c4a0cc5c964ff4dda/larq/quantizers.py#L8-L10

koenhelwegen:
closes #108

### DIFF
--- a/larq/math.py
+++ b/larq/math.py
@@ -4,5 +4,18 @@ import tensorflow as tf
 
 
 def sign(x):
-    """A sign function that will never be zero"""
+    r"""A sign function that will never be zero
+    \\[
+    f(x) = \begin{cases}
+      -1 & x < 0 \\\
+      1 & x \geq 0
+    \end{cases}
+    \\]
+
+    # Arguments
+    `x`: Input `tensor`
+
+    # Returns
+    A `tensor` with same type as `x`.
+    """
     return tf.sign(tf.sign(x) + 0.1)
--- a/larq/math.py
+++ b/larq/math.py
@@ -1,0 +1,8 @@
+"""Math operations that specific to extremely quantized networks."""
+
+import tensorflow as tf
+
+
+def sign(x):
+    """A sign function that will never be zero"""
+    return tf.sign(tf.sign(x) + 0.1)
--- a/larq/math_test.py
+++ b/larq/math_test.py
@@ -4,13 +4,14 @@ import larq as lq
 import pytest
 
 
-@pytest.mark.skipif(not tf.executing_eagerly(), reason="requires eager execution")
 def test_sign():
     # test random
     x = np.random.rand(100) * 100 - 0.5
     y = lq.math.sign(x)
-    np.testing.assert_allclose(y.numpy(), np.array(x >= 0, dtype=np.float32) * 2 - 1)
+    np.testing.assert_allclose(
+        tf.keras.backend.get_value(y), np.array(x >= 0, dtype=np.float32) * 2 - 1
+    )
     # test zero is mapped to one
     np.testing.assert_almost_equal(
-        lq.math.sign(np.zeros(10)).numpy(), np.array(np.ones(10))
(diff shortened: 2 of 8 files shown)


larq/quantizers.py
```diff
@@ -2,31 +2,26 @@
 quantized output and the pseudo-gradient method used for the backwards pass."""
 
 import tensorflow as tf
-from larq import utils
-
-
-def sign(x):
-    """A sign function that will never be zero"""
-    return tf.sign(tf.sign(x) + 0.1)
+import larq as lq
```
lgeiger: I'd prefer to still keep the import like `from larq import utils, math` so users can import `from larq import quantizers` without needing to implicitly import every submodule from larq (due to the `import larq as lq` in this file). I think (though I am not sure) this is what most other libraries to as well.

On the other hand TensorFlow takes ages to import anyway so we definitely don't get any noticable import speed improvements by only importing `utils` and `math` here.

larq/math.py
```diff
@@ -0,0 +1,8 @@
+"""Math operations that specific to extremely quantized networks."""
+
+import tensorflow as tf
+
+
+def sign(x):
+    """A sign function that will never be zero"""
```
lgeiger: I think we should add this to our public documentation, since it can be quite useful.
For this it would be great to add `Arguments` and `Returns` to the docsting.

Also add it here:
https://github.com/plumerai/larq/blob/f713c6ba741dc1c0267a3c0c4a0cc5c964ff4dda/apidocs.yml#L13-L14
and here:
https://github.com/plumerai/larq/blob/f713c6ba741dc1c0267a3c0c4a0cc5c964ff4dda/mkdocs.yml#L12-L19

For it to show up in the build docs.

larq/math_test.py
```diff
@@ -0,0 +1,16 @@
...
+import larq as lq
+import pytest
+
+
+@pytest.mark.skipif(not tf.executing_eagerly(), reason="requires eager execution")
+def test_sign():
+    # test random
+    x = np.random.rand(100) * 100 - 0.5
+    y = lq.math.sign(x)
+    np.testing.assert_allclose(y.numpy(), np.array(x >= 0, dtype=np.float32) * 2 - 1)
```
lgeiger: If you use `tf.keras.backend.get_value(y)`, you can make this test run for both eager and legacy mode.

larq/math_test.py
```diff
@@ -0,0 +1,16 @@
...
+    # test zero is mapped to one
+    np.testing.assert_almost_equal(
+        lq.math.sign(np.zeros(10)).numpy(), np.array(np.ones(10))
```
lgeiger: See above

larq/math_test.py
```diff
@@ -0,0 +1,17 @@
+import tensorflow as tf
+import numpy as np
+import larq as lq
+import pytest
```
lgeiger:
```suggestion
```

larq/optimizers_v2.py
```diff
@@ -76,7 +76,7 @@ def _resource_apply_dense(self, grad, var):
         m_t = tf.compat.v1.assign(
             m, (1 - gamma) * m + gamma * grad, use_locking=self._use_locking
         )
-        var_t = lq.quantizers.sign(-tf.sign(var * m_t - threshold) * var)
+        var_t = lq.math.sign(-tf.sign(var * m_t - threshold) * var)
```
lgeiger: That's a lot more readable now 👍

larq/quantizers.py
```diff
@@ -2,28 +2,24 @@
...
```
```suggestion
```

PR #111, opened 2019-06-25
Add a generic hyperparameter scheduler (closes #104)

### LINKED ISSUE #104: Generic hyperparameter scheduler
The Idea is to have a TF scheduler which is optimizer/hyperparameter agnostic. 
As an example, it can be used for gamma, threshold or learning rate of the floating point optimizer in Bop.

arashb:
this is the first version of the generic HP scheduler.

### DIFF
--- a/larq/callbacks.py
+++ b/larq/callbacks.py
@@ -1,8 +1,6 @@
 import tensorflow as tf
 import numpy as np
 
-from tensorflow import keras
-
 
 class QuantizationLogger(tf.keras.callbacks.Callback):
     """Callback that adds quantization specific metrics.
@@ -81,13 +79,12 @@ class QuantizationLogger(tf.keras.callbacks.Callback):
 class HyperparameterScheduler(tf.keras.callbacks.Callback):
     """Generic hyperparameter scheduler.
 
-  Arguments:
-      schedule: a function that takes an epoch index as input
-          (integer, indexed from 0) and returns a new
-          hyperparameter as output.
-      hyperparameter: str. the name of the hyperparameter to be scheduled.
-      verbose: int. 0: quiet, 1: update messages.
-  """
+    # Arguments:
+    schedule: a function that takes an epoch index as input
+        (integer, indexed from 0) and returns a new hyperparameter as output.
+    hyperparameter: str. the name of the hyperparameter to be scheduled.
+    verbose: int. 0: quiet, 1: update messages.
+    """
 
     def __init__(self, schedule, hyperparameter, verbose=0):
         super(HyperparameterScheduler, self).__init__()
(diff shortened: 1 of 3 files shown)


arashb:
@koenhelwegen @lgeiger this version passes the tests I developed for with TF 1.14+ but before extending it to TF 1.13: 
1. Do you see any possible problem with this approach? 
2. which of the two interfaces makes more sense? to pass the hyperparameter's name as a string or pass the hyperparaters attribue instead of its name?

larq/optimizers_v1.py
```diff
@@ -181,8 +181,8 @@ def __getattr__(self, name):
     def get_config(self):
         fp_optimizer_config = self.fp_optimizer.get_config()
         config = {
-            "threshold": self.threshold,
-            "gamma": self.gamma,
+            "threshold": tf.keras.backend.get_value(self.threshold),
+            "gamma": tf.keras.backend.get_value(self.gamma),
```
lgeiger: Just to be safe, I think we should cast it to a `float` here to be inline with what [keras does](https://github.com/keras-team/keras/blob/master/keras/optimizers.py#L213):
```suggestion
            "gamma": float(tf.keras.backend.get_value(self.gamma)),
```

larq/optimizers_v1.py
```diff
@@ -181,8 +181,8 @@ def __getattr__(self, name):
     def get_config(self):
         fp_optimizer_config = self.fp_optimizer.get_config()
         config = {
-            "threshold": self.threshold,
-            "gamma": self.gamma,
+            "threshold": tf.keras.backend.get_value(self.threshold),
```
lgeiger:
```suggestion
            "threshold": float(tf.keras.backend.get_value(self.threshold)),
```

larq/callbacks_test.py
```diff
@@ -46,3 +52,58 @@ def test_quantization_logger(self):
...
+            x_train.shape[1], 20, y_train.shape[1]
+        )
+        bop_optimizer = lq.optimizers.Bop(fp_optimizer=tf.keras.optimizers.Adam(0.01))
+        model.compile(
+            loss="categorical_crossentropy",
+            optimizer=bop_optimizer,
+            metrics=["accuracy"],
+        )
+
+        scheduler = lambda x: 1.0 / (1.0 + x)
```
lgeiger: I think assigning lambda functions to variables is generally discurraged in Python. Let's just go with a normal function:
```suggestion
        def scheduler(epoch):
            return 1 / (1 + epoch)
```

larq/callbacks.py
```diff
@@ -74,3 +74,46 @@ def on_train_begin(self, logs=None):
...
+            hyperparameter_val = self.schedule(epoch, hyperparameter_val)
+        except TypeError:  # Support for old API for backward compatibility
+            hyperparameter_val = self.schedule(epoch)
+
+        tf.keras.backend.set_value(hp, hyperparameter_val)
+
+        if self.verbose > 0:
+            print(
+                "\nEpoch %05d: %s changning to %s."
+                % (epoch + 1, self.hyperparameter, tf.keras.backend.get_value(hp))
```
lgeiger: I think we can reuse `hyperparameter_val` here, correct?
```suggestion
                % (epoch + 1, self.hyperparameter, hyperparameter_val)
```

larq/callbacks.py
```diff
@@ -74,3 +74,46 @@ def on_train_begin(self, logs=None):
...
+            hyperparameter_val = tf.keras.backend.get_value(hp)
+            hyperparameter_val = self.schedule(epoch, hyperparameter_val)
+        except TypeError:  # Support for old API for backward compatibility
+            hyperparameter_val = self.schedule(epoch)
+
+        tf.keras.backend.set_value(hp, hyperparameter_val)
+
+        if self.verbose > 0:
+            print(
+                "\nEpoch %05d: %s changning to %s."
```
lgeiger: I really like the new f-strings in Python which make this syntax a lot more readable:
```python
print(f"Epoch {epoch + 1}: {self.hyperparameter} changing to {hyperparameter_val}")
```

larq/callbacks.py
arashb: yes, but I wanted to print out the value that is actually stored in optimizer. This is helpful in cases where implicit casting is happening.

lgeiger: I see. For casting it makes sense.

PR #117, opened 2019-07-02
Pin test and doc dependencies

lgeiger:
The upgrade to matplotlib 3.1 broke our CSS on the activations and quantizer pages. Let's pin all the test and doc dependencies so we have a somewhat reproducible environment. @dependabot will assist us with upgrading.


lgeiger:
> The upgrade to matplotlib 3.1 broke our CSS on the activations and quantizer pages

https://github.com/matplotlib/matplotlib/pull/14403 will probably fix this. Matplotlib 3.1.1 should be on PyPi soon, so we should try again then. Let's stick to 3.0.3 for the time being.

PR #119, opened 2019-07-02
Depthwise conv

koenhelwegen:


larq/layers.py
```diff
@@ -1321,3 +1333,46 @@ def __init__(
             implementation=implementation,
             **kwargs
         )
+
+
+@utils.register_keras_custom_object
+class QuantDepthwiseConv2D(QuantizerBase, tf.keras.layers.DepthwiseConv2D):
```
lgeiger: Could you add some docstrings to similar to https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/layers/convolutional.py#L1686-L1747

larq/layers.py
```diff
@@ -1321,3 +1333,46 @@ def __init__(
...
+        self,
+        kernel_size,
+        strides=(1, 1),
+        padding="valid",
+        depth_multiplier=1,
+        data_format=None,
+        activation=None,
+        use_bias=True,
+        input_quantizer=None,
+        kernel_quantizer=None,
```
lgeiger: I think this should be called `depthwise_quantizer` to be consistent with `depthwise_regularizer` and `depthwise_constraint`, `depthwise_initializer`.

larq/layers_test.py
```diff
@@ -101,6 +101,43 @@ def test_binarization(self, quantized_layer, layer, input_shape, kwargs):
...
+    @parameterized.named_parameters(
+        (
+            "QuantDepthwiseConv2D",
+            lq.layers.QuantDepthwiseConv2D,
+            tf.keras.layers.DepthwiseConv2D,
+            (2, 3, 7, 6),
+            dict(kernel_size=3),
+        )
+    )
+    def test_depthwise_layers(self, quantized_layer, layer, input_shape, kwargs):
```
lgeiger: I'd prefer if you either integrate this test into the existing layer tests, or remove the parameterization, since it only has one set of parameters anyway.

larq/layers_test.py
```diff
@@ -165,6 +202,19 @@ def test_layer_does_not_warn(caplog):
     assert caplog.records == []
 
 
+def test_depthwise_layer_warns(caplog):
+    lq.layers.QuantDepthwiseConv2D(5, depthwise_quantizer="ste_sign")
+    assert len(caplog.records) == 1
+    assert "depthwise_constraint" in caplog.text
+
+
+def test_depthwise_layer_does_not_warn(caplog):
```
lgeiger: This is basically the same test as `test_separable_layer_does_not_warn` and `test_layer_does_not_warn`. Could you consolidate them?

larq/layers_test.py
```diff
@@ -197,6 +247,7 @@ def test_separable_layer_does_not_warn(caplog):
         (lq.layers.QuantConv3DTranspose, tf.keras.layers.Conv3DTranspose),
         (lq.layers.QuantLocallyConnected1D, tf.keras.layers.LocallyConnected1D),
         (lq.layers.QuantLocallyConnected2D, tf.keras.layers.LocallyConnected2D),
+        (lq.layers.QuantDepthwiseConv2D, tf.keras.layers.DepthwiseConv2D),
```
lgeiger: :+1:

larq/layers.py
```diff
@@ -76,6 +76,64 @@ def get_config(self):
         return {**super().get_config(), **config}
 
 
+class QuantizerDepthwiseBase(tf.keras.layers.Layer):
+    """Base class for defining quantized layers
+
+    `input_quantizer` and `kernel_quantizer` are the element-wise quantization
```
lgeiger: `kernel_quantizer` --> `depthwise_quantizer`

larq/layers.py
```diff
@@ -76,6 +76,64 @@ def get_config(self):
         return {**super().get_config(), **config}
 
 
+class QuantizerDepthwiseBase(tf.keras.layers.Layer):
```
lgeiger: I would really like to find a better way of moving this logic to a shared base class, since we're basically copy pasting the same code for `QuantizerBase`, `SeperableQuantizerBase` and now `QuantizerDepthwiseBase`. I'd be OK with doing this in a future PR, but as of now this will become a large maintenance burden if we start adding more layers.

koenhelwegen: Agreed. I would argue the logic needs to be implemented at a variable level, but this is tricky to do as it touches a lot of tf internal logic.

larq/layers_test.py
koenhelwegen: I see your point but a merged version would be harder to read and not much shorter. For unit tests, having some duplication is fine I think.

lgeiger: 👍 Sounds good to me

PR #124, opened 2019-07-04
Switch to terminaltables for generating our summary

lgeiger:
This largely reverts #121 and refactors `model.summary()` to use `terminaltabs` for better maintainability in preparation for #91

Here is a example output:
```
+sequential stats-----------------------------------------------------------+
| Layer                            Outputs   # 1-bit  # 32-bit  Memory (kB) |
+---------------------------------------------------------------------------+
| quant_conv2d            (-1, 30, 30, 64)      1728         0         0.21 |
| batch_normalization     (-1, 30, 30, 64)         0       192         0.75 |
| quant_conv2d_1          (-1, 30, 30, 64)     36864         0         4.50 |
| max_pooling2d           (-1, 15, 15, 64)         0         0         0.00 |
| batch_normalization_1   (-1, 15, 15, 64)         0       192         0.75 |
| quant_conv2d_2         (-1, 15, 15, 128)     73728         0         9.00 |
| batch_normalization_2  (-1, 15, 15, 128)         0       384         1.50 |
| quant_conv2d_3         (-1, 15, 15, 128)    147456         0        18.00 |
| max_pooling2d_1          (-1, 7, 7, 128)         0         0         0.00 |
| batch_normalization_3    (-1, 7, 7, 128)         0       384         1.50 |
| quant_conv2d_4           (-1, 7, 7, 256)    294912         0        36.00 |
| batch_normalization_4    (-1, 7, 7, 256)         0       768         3.00 |
| quant_conv2d_5           (-1, 7, 7, 256)    589824         0        72.00 |
| max_pooling2d_2          (-1, 3, 3, 256)         0         0         0.00 |
| batch_normalization_5    (-1, 3, 3, 256)         0       768         3.00 |
| flatten                       (-1, 2304)         0         0         0.00 |
| quant_dense                   (-1, 4096)   9437184         0      1152.00 |
| batch_normalization_6         (-1, 4096)         0     12288        48.00 |
| quant_dense_1                 (-1, 4096)  16777216         0      2048.00 |
| batch_normalization_7         (-1, 4096)         0     12288        48.00 |
| quant_dense_2                   (-1, 10)     40960         0         5.00 |
| batch_normalization_8           (-1, 10)         0        30         0.12 |
| activation                      (-1, 10)         0         0         0.00 |
+---------------------------------------------------------------------------+
| Total                                     27399872     27294      3451.33 |
+---------------------------------------------------------------------------+
+sequential summary----------------+
| Total params           27427166  |
[...]

Issue #91, opened 2019-05-10
Improve model summary


lgeiger:
- Support quantizers not used as a `input_quantizer`
- Support different precisions (e.g. ternary)
- Number of MAC operations
- Memory footprint

jamescook106:
What is wanted in order of priority is:

- [ ] Total memory footprint of the model, and per layer
- [ ] Compression ratio versus Float32 and Float16
- [ ] Support for 8 bit
- [ ] Binary operations and Floating point operations
- [ ] Size of activations
- [ ] Report generator option which is easily readable for post processing
- [ ] Support for 2 bit
- [ ] Support for 4 bit

jamescook106:
@lgeiger anything else?

arashb:
I guess the features requested for model summary are overlapping with the profiler tool I will be working on or we just combine the two tasks?

lgeiger:
Closing in favour of #137

PR #132, opened 2019-07-09
Add support for arbitrary precision in model summary

lgeiger:
This will add support for arbitrary precision in model summary and correctly list ternary and 8-bit layers in the summary. This doesn't change the output introduced in #124 but will add support for quantizers with arbitrary precisions.

Each `quantizer` will now have an additional `precision` property that is used to determine the expected output precision and is parsed by the model summary. This is a very ad-hock approach to adding an improved summary, but is definitely an improvement on the current status.

Adresses parts of #91


jamescook106:
Please post an example output

lgeiger:
> Please post an example output

It is the same as in #124

For a table with ternary a network checkout the snapshot test [here](https://github.com/plumerai/larq/pull/132/files#diff-2c2775c6c10e7a2fdca7d69095aa0a73)

lgeiger:
Here is a example output from a toy-network with different precisions:
```
+sequential_1 stats----------------------------------------------------------+
| Layer                     Outputs  # 1-bit  # 2-bit  # 32-bit  Memory (kB) |
+----------------------------------------------------------------------------+
| quant_conv2d_1   (-1, 26, 26, 32)      288        0        32         0.16 |
| max_pooling2d_1  (-1, 13, 13, 32)        0        0         0         0.00 |
| quant_conv2d_2   (-1, 11, 11, 32)        0     9216        32         2.38 |
| dense_1          (-1, 11, 11, 10)        0        0       330         1.29 |
+----------------------------------------------------------------------------+
| Total                                  288     9216       394         3.82 |
+----------------------------------------------------------------------------+
+sequential_1 summary------------+
| Total params           9898    |
| Trainable params       9898    |
| Non-trainable params   0       |
| Float-32 Equivalent    0.04 MB |
| Compression of Memory  10.11   |
+--------------------------------+
```

PR #136, opened 2019-07-10
Ignore metric weights in layers

lgeiger:
This will ignore the weights added by metrics when calling `layer.weights`. This has a few benefits:
- Weights from layer metrics will not be saved when calling `model`. This means that:
  - checkpoints are compatible between different versions of `larq` and when disabling metrics on the user side. I discovered this when trying to reload an old weights file in `larq-zoo` that didn't include the metrics.
  - checkpoints are smaller since they don't include unnecessary items
- We don't need to handle the ignoring in the `model.summary` code (which is already quite bloated)
- In general interacting with `layer.weights` will be easier since weights related to metrics are hidden for users

### DIFF
--- a/larq/layers.py
+++ b/larq/layers.py
@@ -56,6 +56,14 @@ class QuantizerBase(tf.keras.layers.Layer):
                     values_shape=self.kernel.shape, name=f"{self.name}/flip_ratio"
                 )
 
+    @property
+    def non_trainable_weights(self):
+        weights = super().non_trainable_weights
+        if hasattr(self, "flip_ratio"):
+            metrics_weights = self.flip_ratio.weights
+            return [weight for weight in weights if weight not in metrics_weights]
+        return weights
+
     def call(self, inputs):
         if self.input_quantizer:
             inputs = self.input_quantizer(inputs)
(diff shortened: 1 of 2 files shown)


larq/layers.py
```diff
@@ -56,6 +56,14 @@ def build(self, input_shape):
                     values_shape=self.kernel.shape, name=f"{self.name}/flip_ratio"
                 )
 
+    @property
+    def non_trainable_weights(self):
```
lgeiger: We really need to find a better way to share logic between layers. I'll do some refactoring next week. We should also checkout how TensorFlow is handling their automatic mixed precision [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/mixed_precision/experimental/autocast_variable.py) to maybe gain some inspiration.

lgeiger: @koenhelwegen Do you think messing with `non_trainable_weights` could have any unexpected side effects?

I also tried setting `non_trainable_weights = []` on the metrics layer directly, but that doesn't work since initialisation wouldn't be handled correctly.

koenhelwegen: I have a hard time judging if this is save. But I don't really understand why we have to mess with this in the first place. TF Keras metrics are also stateful, so is there not a more standard way of dealing with this?

lgeiger: > TF Keras metrics are also stateful, so is there not a more standard way of dealing with this?

TF Keras metrics are added to the model. We add our metrics to the layer since we want to observe changes in the forward pass. Underneath a metric is a normal Keras layer and therefor adding weights will add them to the layer as well.
`layer.weights` internally calls `layers.non_trainable_weights` and `layers.trainable_weights` and will also get all the weights from its child layers (which is important if you have layers that bundle multiple layers). I checked, that the metrics are correctly updated during training in TF 1.14 and TF 2. It seams that we can safely make this change without messing things up.

I'll open an issue on the TensorFlow repo on Monday to see if this behaviour is intended or an oversight. The `add_metrics` API is quite new so there aren't a lot of examples out there that use this on a layer level.

koenhelwegen: okay, thanks! I will merge it and if we find a better way we can change it later

PR #143, opened 2019-07-12
Fetch larq-zoo docs before running mkdocs

lgeiger:
This will add a Model tab on our site, that will include the docs for `larq-zoo`

### DIFF
--- a/larqdocs.sh
+++ b/larqdocs.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -e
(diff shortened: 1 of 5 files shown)


larqdocs.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
lgeiger: In the future we should probably write a proper mkdocs plugin for this. For now, this will work

PR #144, opened 2019-07-15
Correct inequality sign in SteTern documentation

MariaHeuss:
Changed previously wrong direction of one inequality symbol in SteTern documentation

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -137,7 +137,7 @@ class SteTern:
     q(x) = \begin{cases}
     +1 & x > \Delta \\\
     0 & |x| < \Delta \\\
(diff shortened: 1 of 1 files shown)


larq/quantizers.py
```diff
@@ -136,7 +136,7 @@ class SteTern:
     \\[
     q(x) = \begin{cases}
     +1 & x > \Delta \\\
-    0 & |x| > \Delta \\\
+    0 & |x| < \Delta \\\
     -1 & x < \Delta
```
lgeiger: This line needs to be changed too, correct?
```suggestion
     -1 & x < - \Delta
```

PR #146, opened 2019-07-15
Improved hyperlinks, plus docs on quantizers as activations

### LINKED ISSUE #92: Document how to use a quantizer as an activation function
`tf.keras.layers.Activation("ste_sign")`

koenhelwegen:
closes #92


docs/guides/key-concepts.md
```diff
@@ -12,6 +12,26 @@ It is also common to apply quantizers to the weights during training. This is ne
...
+
+Although quantizers are usually passed to specialized arguments (see [Quantized Layers](#quantized-layers)), they can be used just like `tf.keras` activations:
+
+```python
+# Use a quantizer als activation
+y = larq.layers.QuantDense(
+        512,
+        activation="ste_sign",
+        )(x)
+```
```
lgeiger: Does this fit into one line? I think it looks a bit weird like this.

docs/guides/key-concepts.md
```diff
@@ -12,6 +12,26 @@ It is also common to apply quantizers to the weights during training. This is ne
 
 The true gradient of a quantizer is in general zero almost everywhere and therefore cannot be used for gradient descent. Instead, optimization of BNNs rely on what we call pseudo-gradients, which are used during back-propagation. In the documentation for each quantizer you will find the definition and a graph of the pseudo-gradient.
 
+### Using Quantizers as Activations
```
lgeiger: Do you think it would be good to add a similar section [here](https://github.com/larq/larq/blob/master/larq/quantizers.py#L1) and [here](https://github.com/larq/larq/blob/master/larq/layers.py#L1) too? Very much like we do it for the [activations](https://github.com/larq/larq/blob/master/larq/activations.py#L1)

docs/guides/key-concepts.md
```diff
@@ -12,6 +12,26 @@ It is also common to apply quantizers to the weights during training. This is ne
...
+        activation="ste_sign",
+        )(x)
+```
+
+Just like activations, quantizers can also be used on their own:
+
+```python
+# The two lines below are equivalent
+x_binarized = larq.quantizers.ste_sign(x)
+x_binarized = tf.keras.layers.Activation("ste_sign")(x)
```
lgeiger: Separate from this PR do you think we should add a class that could be used like this?
```python
larq.layers.Quantizer("ste_sign")(x)
```

docs/guides/key-concepts.md
koenhelwegen: I added a section to quantizers.py, but what exactly would you like to see in layers.py?

lgeiger: Yeah, you are right. I though it would make sense to maybe explain the `input_quantizer` in the layers file too, since it is a layer kwarg.

PR #151, opened 2019-07-17
Add note on ternarization threshold

MariaHeuss:
Added a note but I am uncertain whether we should also change the default parameter of ternary_weight_networks to True.

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -181,9 +181,7 @@ class SteTern:
     \\[
     \Delta = \frac{0.7}{n} \sum_{i=1}^{n} |W_i|
     \\]
-    where we assume that $W_i$ is generated from a normal distribution. Note that a too high threshold may 
-    cause the ternarization to set most input values to zero and thereby prevent the network from learning. Thererfore 
-    we advise against using the default parameter and rather recommend using the given calcuation.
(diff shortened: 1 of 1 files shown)


larq/quantizers.py
```diff
@@ -181,7 +181,9 @@ class SteTern:
     \\[
     \Delta = \frac{0.7}{n} \sum_{i=1}^{n} |W_i|
     \\]
-    where we assume that $W_i$ is generated from a normal distribution.
+    where we assume that $W_i$ is generated from a normal distribution. Note that a too high threshold may 
+    cause the ternarization to set most input values to zero and thereby prevent the network from learning. Thererfore 
+    we advise against using the default parameter and rather recommend using the given calcuation.
```
lgeiger: If possible I think we should change the default threshold to a more sensible value, rather than adding a note that advises users to change the value themselves.

jamescook106: Yeah if we turn the default value to 0.05 that should be good enough. I would advise not using the paper calculation for what the value should be unless it has been commonly used in papers since, and can be used for the activations also

MariaHeuss: Ah sorry I just saw this now after the second commit. It is commonly used in papers although I am uncertain about whether it is also used for the activations. I can anyways just set the default lower if you think this is better.

PR #157, opened 2019-07-19
refactoring plus mac-counter

koenhelwegen:
Draft because I still need to do the testing. Example output:

```
+model stats-------------------------------------------------------------------------------------------------------+
| Layer                  Input prec.                 Outputs   # 1-bit  # 32-bit   Memory  1-bit macs  32-bit macs |
|                              (bit)                               x 1       x 1     (kB)        (kB)         (kB) |
+------------------------------------------------------------------------------------------------------------------+
| input_1                          -  ((None, 224, 224, 3),)         0         0        0           ?            ? |
| quant_conv2d                     -        (-1, 56, 56, 64)     23232        64     3.09           0     158.8125 |
| max_pooling2d                    -        (-1, 27, 27, 64)         0         0        0           0            0 |
| batch_normalization              -        (-1, 27, 27, 64)         0       192     0.75           0            0 |
| quant_conv2d_1                   1       (-1, 27, 27, 192)    307200       192    38.25      1012.5            0 |
| max_pooling2d_1                  -       (-1, 13, 13, 192)         0         0        0           0            0 |
| batch_normalization_1            -       (-1, 13, 13, 192)         0       576     2.25           0            0 |
| quant_conv2d_2                   1       (-1, 13, 13, 384)    663552       384    82.50      1053.0            0 |
| batch_normalization_2            -       (-1, 13, 13, 384)         0      1152     4.50           0            0 |
| quant_conv2d_3                   1       (-1, 13, 13, 384)   1327104       384   163.50      2106.0            0 |
| batch_normalization_3            -       (-1, 13, 13, 384)         0      1152     4.50           0            0 |
| quant_conv2d_4                   1       (-1, 13, 13, 256)    884736       256   109.00      1404.0            0 |
| max_pooling2d_2                  -         (-1, 6, 6, 256)         0         0        0           0            0 |
| batch_normalization_4            -         (-1, 6, 6, 256)         0       768     3.00           0            0 |
| flatten                          -              (-1, 9216)         0         0        0           0            0 |
| quant_dense                      1              (-1, 4096)  37748736         0  4608.00     4608.00            0 |
| batch_normalization_5            -              (-1, 4096)         0     12288    48.00           0            0 |
| quant_dense_1                    1              (-1, 4096)  16777216         0  2048.00     2048.00            0 |
| batch_normalization_6            -              (-1, 4096)         0     12288    48.00           0            0 |
| quant_dense_2                    1                (-1, 10)     40960         0     5.00        5.00            0 |
| batch_normalization_7            -                (-1, 10)         0        30     0.12           0            0 |
| activation                       -                (-1, 10)         0         0        0           ?            ? |
+------------------------------------------------------------------------------------------------------------------+
| Total                                                       57772736     29726  7168.45     12236.5     158.8125 |
+------------------------------------------------------------------------------------------------------------------+
+model summary--------------------------------+
| Total params                      57802462  |
| Trainable params                  57783498  |
| Non-trainable params              18964     |
| Model size:                       7.00 MB   |
| Float-32 Equivalent               220.50 MB |
| Compression Ratio of Memory       0.03      |
| Number of MACs                    101542400 |
| Ratio of MACs that are binarized  0.9872    |
+---------------------------------------------+
```


larq/models.py
```diff
@@ -47,82 +64,286 @@ def _get_output_shape(layer):
...
+        if len(self.output_shape) == 4:
+            return np.prod(self.output_shape[1:2])
+        elif len(self.output_shape) == 2:
+            return 1
+        else:
+            raise NotImplementedError()
+
+    @property
+    def unique_param_bidtwidths(self):
+        return sorted(np.unique([p.bitwidth for p in self.parameter_profiles]))
```
lgeiger: Since we are sorting a pure Python `list`, converting it to a `set` is usually faster than using `np.unique` since this would need to convert it to an `np.ndarray` and then back to be used by `sorted`.
```suggestion
        return sorted(set([p.bitwidth for p in self.parameter_profiles]))
```

larq/models.py
```diff
@@ -47,82 +64,286 @@ def _get_output_shape(layer):
...
+
+    @property
+    def unique_op_precisions(self):
+        return sorted(np.unique([op.precision for op in self.op_profiles]))
```
lgeiger:
```suggestion
        return sorted(set([op.precision for op in self.op_profiles]))
```

larq/models.py
```diff
@@ -47,82 +64,286 @@ def _get_output_shape(layer):
...
+            [l.parameter_count(bitwidth, trainable) for l in self.layer_profiles]
+        )
+
+    def op_count(self, bitwidth=None, op_type=None):
+        return sum([l.op_count(bitwidth, op_type, 0) for l in self.layer_profiles])
+
+    @property
+    def unique_param_bidtwidths(self):
+        return sorted(
+            np.unique(
```
lgeiger:
```suggestion
            set(
```

larq/models.py
```diff
@@ -47,82 +64,286 @@ def _get_output_shape(layer):
...
+                _flatten([l.unique_param_bidtwidths for l in self.layer_profiles])
+            )
+        )
+
+    @property
+    def unique_op_precisions(self):
+        return sorted(
+            np.unique(_flatten([l.unique_op_precisions for l in self.layer_profiles]))
```
lgeiger:
```suggestion
            set(_flatten([l.unique_op_precisions for l in self.layer_profiles]))
```

larq/models.py
```diff
@@ -47,82 +64,286 @@ def _get_output_shape(layer):
...
+            return 32
+
+
+class ModelProfile:
+    def __init__(self, model):
+        self.layer_profiles = [LayerProfile(l) for l in model.layers]
+
+    @property
+    def memory(self):
+        return sum([l.memory for l in self.layer_profiles])
```
lgeiger:
```suggestion
        return sum(l.memory for l in self.layer_profiles)
```

larq/models.py
```diff
@@ -47,82 +64,286 @@ def _get_output_shape(layer):
...
+
+    @property
+    def fp_equivalent_memory(self):
+        return sum([l.fp_equivalent_memory for l in self.layer_profiles])
```
lgeiger:
```suggestion
        return sum(l.fp_equivalent_memory for l in self.layer_profiles)
```

larq/models.py
```diff
@@ -47,82 +64,286 @@ def _get_output_shape(layer):
...
+
+    def parameter_count(self, bitwidth=None, trainable=None):
+        return sum(
+            [l.parameter_count(bitwidth, trainable) for l in self.layer_profiles]
```
lgeiger:
```suggestion
            l.parameter_count(bitwidth, trainable) for l in self.layer_profiles
```

larq/models.py
```diff
@@ -47,82 +64,286 @@ def _get_output_shape(layer):
...
+        )
+
+    def op_count(self, bitwidth=None, op_type=None):
+        return sum([l.op_count(bitwidth, op_type, 0) for l in self.layer_profiles])
```
lgeiger:
```suggestion
        return sum(l.op_count(bitwidth, op_type, 0) for l in self.layer_profiles)
```

larq/models.py
```diff
@@ -47,82 +64,286 @@ def _get_output_shape(layer):
...
+
+    def _get_bitwidth(self, weight):
+        try:
+            quantizer = self._layer.quantizers[self._quantized_weights().index(weight)]
+            return quantizer.precision
+        except:
+            return 32
+
+
+class ModelProfile:
```
lgeiger: Does it make sense to inherit from `LayerProfile` here?

jamescook106:
Would you be able to post an example output for Ternary?

larq/models.py
```diff
@@ -47,82 +64,286 @@ def _get_output_shape(layer):
...
+    def fp_equivalent_memory(self):
+        return sum([p.fp_equivalent_memory for p in self.parameter_profiles])
+
+    def parameter_count(self, bitwidth=None, trainable=None):
+        count = 0
+        for p in self.parameter_profiles:
+            if bitwidth and (p.bitwidth != bitwidth):
+                continue
+            if (trainable is not None) and (p.trainable != trainable):
+                continue
```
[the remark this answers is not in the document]
koenhelwegen: me neither, but I don't have anything cleaner atm

larq/models.py
koenhelwegen: Currently I don't see a way to do this nicely. Eventually it would be good to support nested models.

Interesting to see how Keras will handle this: https://github.com/tensorflow/tensorflow/issues/28154

koenhelwegen:
> Would you be able to post an example output for Ternary?

There is a ternary layer in the profile used for the test :)

larq/models.py
```diff
@@ -1,41 +1,61 @@
...
+    keras_layers.SeparableConv2D,
+    keras_layers.DepthwiseConv2D,
+    keras_layers.Dense,
+    keras_layers.Flatten,
+    keras_layers.BatchNormalization,
+    keras_layers.MaxPool2D,
+    keras_layers.AveragePooling2D,
+]
+
+mac_layers = [
```
MariaHeuss: I would name it mac_containing_layers

larq/models.py
```diff
@@ -47,91 +67,316 @@ def _get_output_shape(layer):
...
+    def __init__(self, layer):
+        self._layer = layer
+        self.parameter_profiles = [
+            _parameterProfile(weight, self._get_bitwidth(weight))
+            for weight in layer.weights
+        ]
+
+        self.op_profiles = []
+
+        if type(layer) in mac_layers:
```
lgeiger: Does this do the right thing?

I think if `mac_layers` would be a tuple you can could use:
```python
if isinstance(layer, mac_layers):
```
Not sure if it works with a `list` too.

larq/models.py
```diff
@@ -47,91 +67,316 @@ def _get_output_shape(layer):
...
+                trainable is None or p.trainable == trainable
+            ):
+                count += p.count
+        return count
+
+    def op_count(self, op_type=None, precision=None, escape="?"):
+        if op_type != "mac":
+            raise ValueError("Currently only counting of MAC-operations is supported.")
+
+        if type(self._layer) in op_count_supported_layer_types:
```
lgeiger: See comment above

larq/models.py
lgeiger: Sorry for the confusion, I just ment that it would make sense to inherit `ModelProfile` from `LayerProfile` so you don't have to redefine `memory` and `fp_equivalent_memory`.

larq/models.py
```diff
@@ -1,41 +1,61 @@
...
-    weights: An iterable containing the weights on which to compute params
 
-    # Returns
-    The total number of scalars composing the weights
-    """
-    return int(sum(np.prod(w.shape.as_list()) for w in weights))
+def _format_table_entry(x, units=1):
+    try:
+        assert not np.isnan(x)
+        if type(x) == str or x == 0 or units == 1:
```
MariaHeuss: would maybe "if type(x) == str: ..." be better because a division is faster than an if statement?

larq/models.py
```diff
@@ -47,91 +67,316 @@ def _get_output_shape(layer):
...
+    def parameter_count(self, bitwidth=None, trainable=None):
+        return sum(l.parameter_count(bitwidth, trainable) for l in self.layer_profiles)
+
+    def op_count(self, op_type=None, bitwidth=None):
+        return sum(l.op_count(op_type, bitwidth, 0) for l in self.layer_profiles)
+
+    @property
+    def unique_param_bidtwidths(self):
+        return sorted(
+            set(_flatten([l.unique_param_bidtwidths for l in self.layer_profiles]))
```
lgeiger: Does your `_flatten` also accept a generator expression?  If so I'd remove the `[` and `]`.

larq/models.py
```diff
@@ -47,91 +67,316 @@ def _get_output_shape(layer):
...
             "Outputs",
-            *(f"# {i}-bit" for i in sorted(summed_stat)),
-            "Memory\n(kB)",
+            *(
+                f"# {i}-bit\nx {table_config['param_units']}"
+                for i in table_config["param_bidtwidths"]
+            ),
+            f"Memory\n({_bitsize_as_str(table_config['memory_units'])})",
+            *(
+                f"{i}-bit macs\n({_bitsize_as_str(table_config['mac_units'])})"
```
lgeiger: What do you think about
```suggestion
                f"{i}-bit MACs\n({_bitsize_as_str(table_config['mac_units'])})"
```

larq/models.py
```diff
@@ -47,91 +67,316 @@ def _get_output_shape(layer):
...
         ]
-    )
-    return table
+
+        if include_macs:
+            binarization_ratio = self.op_count("mac", 1) / self.op_count(op_type="mac")
+            summary.extend(
+                [
+                    ["Number of MACs", self.op_count(op_type="mac")],
+                    ["Ratio of MACs that are binarized", f"{binarization_ratio:.4f}"],
```
lgeiger: What do you think about
```suggestion
                    ["Ratio of binarized MACs", f"{binarization_ratio:.4f}"],
```

larq/snapshots/snap_models_test.py
```diff
@@ -7,23 +7,30 @@
...
++sequential summary-------------------------+
+| Total params                      40682   |
+| Trainable params                  40682   |
+| Non-trainable params              0       |
+| Model size:                       0.15 MB |
+| Float-32 Equivalent               0.16 MB |
+| Compression Ratio of Memory       0.96    |
+| Number of MACs                    1411968 |
+| Ratio of MACs that are binarized  0.1124  |
++-------------------------------------------+
```
jamescook106: For me the Ratio of MACS that are binarized makes sense for binary networks, however here we have also ternary, so maybe we could add another line which is the Ratio of MACS that are ternarized

larq/models.py
```diff
@@ -47,91 +67,316 @@ def _get_output_shape(layer):
...
+            for weight in layer.weights
+        ]
+
+        self.op_profiles = []
+
+        if type(layer) in mac_layers:
+            # mac layers should have a kernel weight and perhaps a bias
+            # assert len(self.parameter_profiles) <= 2 \
+            #   or ('Seperable' in str(type(layer)) and len(self.parameter_profiles) <= 3)
+
```
MariaHeuss: remove this.

larq/models.py
koenhelwegen: Would be nice, but this requires to rename the `parameter_profiles` to `profiles` in `LayerProfile`, which I don't think is a good idea.

larq/models.py
koenhelwegen: They `x==0` is there because otherwise `0` is converted to float and printed as `0.00`.  Similarly, `type(2/1.) == float` while `1 == 1.`, so escaping `units == 1` avoids ints unnecessarily converting to floats.

PR #160, opened 2019-07-24
Automatically publish docs from CI

lgeiger:
This will remove the need to publish the docs manually.


lgeiger:
> Do you really want to deploy the docs after each merge to master?

That's how we currently do it manually anyway.

PR #162, opened 2019-07-25
Add sticky tab navigation

lgeiger:
I used my rusty HTML/CSS skills, to play a bit with our documentation site:
This changes the top level navigation have a persistent tab bar:
Bildschirmfoto 2019-07-26 um 00 07 24

One thing I am not sure about is if it is good to repeat the top-level navigation on the left sidebar or should we switch that back to how it is currently done in https://larq.dev/?

### DIFF
--- a/docs-theme/partials/header.html
+++ b/docs-theme/partials/header.html
@@ -1,0 +1,47 @@
+{#-
+  This file was automatically generated - do not edit
+-#}
+<header class="md-header" data-md-component="header">
+  <nav class="md-header-nav md-grid">
+    <div class="md-flex">
+      <div class="md-flex__cell md-flex__cell--shrink">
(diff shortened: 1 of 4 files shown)


MariaHeuss:
I have a week preference on having less information/text on one site, therefore I would go with the persistent tab bar but the side bar as it is currently done. But as I said no strong preference.

koenhelwegen:
Agree with Maria, the sticky top bar is great but would keep left panel as it is now.

lgeiger:
OK, I'll take a look and see if I can figure it out how to make it work.

lgeiger:
I simplified the side navigation to not include the top level:
Bildschirmfoto 2019-07-26 um 14 48 29

PR #163, opened 2019-07-26
Add metrics scope contextmanager

lgeiger:
This allows us to configure training metrics using a context manager similar to `tf.keras.utils.custom_object_scope` or `tf.name_scope`.

I disabled the metrics by default since they can slightly increase runtime and memory consumption.

The API I propose is the following:
```python
with lq.metrics.scope(["flip_ratio"]):
    model = tf.keras.models.Sequential([lq.layers.QuantDense(3, kernel_quantizer="ste_sign", input_shape=(32,))])
model.compile(loss="mse", optimizer="sgd")
```

I still need to add documentation, but I though it would be good to get some feedback first.

### DIFF
--- a/larq/layers.py
+++ b/larq/layers.py
@@ -75,7 +75,8 @@ class QuantDense(QuantizerBase, tf.keras.layers.Dense):
         the output of the layer (its "activation").
     kernel_constraint: Constraint function applied to the `kernel` weights matrix.
     bias_constraint: Constraint function applied to the bias vector.
-    metrics: An array of metrics to add to the layer. Available metrics: `flip_ration`.
+    metrics: An array of metrics to add to the layer. If `None` the metrics set in
+        `larq.metrics.scope` are used. Available metrics: `flip_ration`.
 
     # Input shape
     N-D tensor with shape: `(batch_size, ..., input_dim)`. The most common situation
@@ -168,7 +169,8 @@ class QuantConv1D(QuantizerBase, tf.keras.layers.Conv1D):
         the output of the layer (its "activation").
     kernel_constraint: Constraint function applied to the kernel matrix.
     bias_constraint: Constraint function applied to the bias vector.
-    metrics: An array of metrics to add to the layer. Available metrics: `flip_ration`.
+    metrics: An array of metrics to add to the layer. If `None` the metrics set in
+        `larq.metrics.scope` are used. Available metrics: `flip_ration`.
 
     # Input shape
     3D tensor with shape: `(batch_size, steps, input_dim)`
@@ -274,7 +276,8 @@ class QuantConv2D(QuantizerBase, tf.keras.layers.Conv2D):
         the output of the layer (its "activation").
     kernel_constraint: Constraint function applied to the kernel matrix.
     bias_constraint: Constraint function applied to the bias vector.
-    metrics: An array of metrics to add to the layer. Available metrics: `flip_ration`.
+    metrics: An array of metrics to add to the layer. If `None` the metrics set in
+        `larq.metrics.scope` are used. Available metrics: `flip_ration`.
 
     # Input shape
     4D tensor with shape:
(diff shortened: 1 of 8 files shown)


koenhelwegen:
Awesome! I like this much better. In your example, does the `model.compile()` statement needs to be within the context scope? And the old API is still available right?

lgeiger:
> In your example, does the `model.compile()` statement needs to be within the context scope?

Good point. It is not necessary. I updated the unit test and example above.

> And the old API is still available right?

I intend to remove the metrics `kwarg` since it was not documented and only intended as a internal escape hatch if we run into unexpected issues.

MariaHeuss:
It looks good but I would not put too much weight on my opinion for this one.

koenhelwegen:
I would be in favour of turning the `metrics` kwarg into a permanent, documented feature. It may very well be that a researcher is interested in monitoring only one or two layers, especially as we add more custom metrics. Then a context manager is a bit awkward. This is also what I'm used to from TF (i.e. using a context manager to set a default, and where needed overwriting that with kwargs).

But it adds additional code surface and TF itself is moving away from this coding style a feel, so I'm also fine with not having it.

lgeiger:
> I would be in favour of turning the metrics kwarg into a permanent, documented feature. It may very well be that a researcher is interested in monitoring only one or two layers, especially as we add more custom metrics. Then a context manager is a bit awkward. This is also what I'm used to from TF (i.e. using a context manager to set a default, and where needed overwriting that with kwargs).

After some thinking I agree, that this is a useful feature.

I updated the PR and added documentation. This is ready for a thorough review.

larq/layers.py
```diff
@@ -75,6 +75,8 @@ class QuantDense(QuantizerBase, tf.keras.layers.Dense):
         the output of the layer (its "activation").
     kernel_constraint: Constraint function applied to the `kernel` weights matrix.
     bias_constraint: Constraint function applied to the bias vector.
+    metrics: An array of metrics to add to the layer. If `None` the metrics set in
+        `larq.metrics.scope` are used. Available metrics: `flip_ration`.
```
koenhelwegen:
```suggestion
        `larq.metrics.scope` are used. Available metrics: `flip_ratio`.
```

Would it make sense to make this a hyperlink?

larq/layers.py
```diff
@@ -165,6 +169,8 @@ class QuantConv1D(QuantizerBase, tf.keras.layers.Conv1D):
         the output of the layer (its "activation").
     kernel_constraint: Constraint function applied to the kernel matrix.
     bias_constraint: Constraint function applied to the bias vector.
+    metrics: An array of metrics to add to the layer. If `None` the metrics set in
+        `larq.metrics.scope` are used. Available metrics: `flip_ration`.
```
koenhelwegen:
```suggestion
        `larq.metrics.scope` are used. Available metrics: `flip_ratio`.
```

larq/layers.py
```diff
@@ -268,6 +276,8 @@ class QuantConv2D(QuantizerBase, tf.keras.layers.Conv2D):
...
```
koenhelwegen:
```suggestion
        `larq.metrics.scope` are used. Available metrics: `flip_ratio`.
```

larq/layers.py
```diff
@@ -378,6 +390,8 @@ class QuantConv3D(QuantizerBase, tf.keras.layers.Conv3D):
...
```
koenhelwegen:
```suggestion
        `larq.metrics.scope` are used. Available metrics: `flip_ratio`.
```

koenhelwegen: I will leave the rest for a search replace ;)

larq/layers_test.py
```diff
@@ -77,16 +77,17 @@ def test_binarization(self, quantized_layer, layer, input_shape, kwargs):
...
-            quantized_layer,
-            kwargs=dict(
-                **kwargs,
-                kernel_quantizer="ste_sign",
-                input_quantizer="ste_sign",
-                kernel_initializer=tf.keras.initializers.constant(random_weight),
-            ),
-            input_data=input_data,
-        )
+        with lq.metrics.scope(["flip_ratio"]):
```
koenhelwegen: I do think we should have unit test for the `metrics` kwarg as well

larq/layers.py
lgeiger: Currently we don't have a documentation page for `flip_ratio` since I am still unsure if we want to change the API of the `FlipRatio` class in the future. I think it is fine to keep it as is and we can add more documentation on metrics when we add more.

larq/layers.py
lgeiger: Good catch

larq/layers_test.py
lgeiger: I added a unit test in the latest commit

PR #171, opened 2019-08-06
Cleanup ternary quantizer and add alias to ste_tern

lgeiger:
This fixes the documentation and adds an alias to `ste_tern` to make it consistent with other quantizers.

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -247,6 +247,7 @@ class SteTern:
             "ternary_weight_networks": self.ternary_weight_networks,
         }
 
+
 ste_tern = SteTern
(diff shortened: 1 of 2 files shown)


MariaHeuss:
Nice, I like this better. I was thinking about whether changing the name of the boolean value 'ternary_weight_networks' to 'calculate_threshold', or something similar would be a bit more descriptive. What do you think?

lgeiger:
I'd be happy to change the naming. Though this would be a breaking change, so I think we should do this in a separate PR then.

MariaHeuss:
Yes, true.

PR #172, opened 2019-08-06
Add SwishSign quantizer

lgeiger:
I added this quantizer some time ago, but never pushed it. This adds the swish sign quantizer from https://arxiv.org/abs/1812.11800

Note that this doesn't implement a trainable beta. So we might not want to merge this right away, but wait until I find time to implement a learnable beta parameter which might change the API of this function.

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -63,6 +63,8 @@ __all__ = [
     "approx_sign",
     "MagnitudeAwareSign",
     "magnitude_aware_sign",
+    "swish_sign",
+    "SwishSign",
     "SteTern",
     "ste_tern",
 ]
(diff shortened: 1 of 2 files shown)


jamescook106:
@lgeiger tests are failing

lgeiger:
It looks like class serialization in TF 1.13 is broken. I am skipping this test for now.

I also tried using a trainable beta, but the naive implementation doesn't correctly get the gradient. I only skimmed the paper, so I am not sure if there is a formula mentioned there. We can look into that once we train some models with this quantizer.

lgeiger:
> Is it worth adding in the documentation that beta here is non learnable, whereas in the paper it is?

I think it is fine since all the ImageNet experiments in the paper don't use a learnable beta either

jamescook106:
Please rebase this, and I will re-review and merge

lgeiger:
I will push a rebased version once #182 is merged or closed, since this will introduce more conflicts.

MariaHeuss:
I will review once the rebased version is pushed.

lgeiger:
This is now rebased and ready for review.

lgeiger:
I agree, that having a unit test for the backwards pass would be great. Though since the function for the backwards pass is not trivial (and I would need to implement it again in numpy any way), I tried to keep it simple and just compared the plot in the docs with the figure in the paper to check that it is correct.

Do you think that is enough? Otherwise I can add a unit test tomorrow.

> it may be nice to have an integration test for the backward pass, e.g. by running something like our optimizer test for each quantizer. But that can be done in separate PR.

Our docs build exercises the backwards pass for every quantizer, so we already have a sanity check that way.

koenhelwegen:
Fair point. I do think it would be better to have the test, if only for consistency (for the other quantizers it's also just a numpy reimplementation), but I'm fine either way.

lgeiger:
> Fair point. I do think it would be better to have the test, if only for consistency (for the other quantizers it's also just a numpy reimplementation), but I'm fine either way.

Sorry for being lazy. I added a unit test for the gradient in d952063

PR #178, opened 2019-08-12
Properly register aliases and convert weight_clip to dataclass

lgeiger:
This adds a new `utils.register_alias` function that can be used to correctly register aliases. It furthermore simplifies the current `WeightClip` constrain by converting it to a `dataclass`.

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -248,7 +248,7 @@ class SteTern:
         }
 
 
-ste_tern = SteTern
+ste_tern = utils.register_alias(SteTern, "ste_tern")
 
 
 def serialize(initializer):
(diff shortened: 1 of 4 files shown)


larq/quantizers.py
```diff
@@ -248,7 +248,7 @@ def get_config(self):
         }
 
 
-ste_tern = SteTern
+ste_tern = utils.register_alias(SteTern, "ste_tern")
```
MariaHeuss: I don't really like that ste_tern is an alias for SteTern. It is different in usage from ste_sign but intuitively it should be the same. My proposal is that ste_tern is some default ternarizer such as SteTern(). Anyways this might be another issue again and I can change that later.

lgeiger: Good point. I will change this tomorrow, since this requires some more changes.

PR #181, opened 2019-08-13
Cleanup quantizer classes

lgeiger:
This adds `QuantizerFunctionWrapper` that can be used to wrap a simple quantizer function into a serializable class that will retain the config after reloading from a `.h5` model file.

This allows quantizers to be used either as either
```python
input_quantizer=lq.quantizers.ste_tern
```
or
```python
input_quantizer=lq.quantizers.SteTern(threshold_value=0.05)
```

As mentioned in https://github.com/larq/larq/pull/178#discussion_r312930868 the current `ste_tern` alias is a bit inconsistent this PR cleans this up and introduces a proper way to add configurable quantizers.

### DIFF
--- a/apidocs.yml
+++ b/apidocs.yml
@@ -18,8 +18,4 @@
 - metrics.md:
     - larq.metrics+
 - quantizers.md:
-    - larq.quantizers:
-        - larq.quantizers.ste_sign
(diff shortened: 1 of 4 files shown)


lgeiger:
This PR is now ready for review and supports TF 1.13 as well

PR #182, opened 2019-08-13
Allow to configure gradient clipping in quantizers

lgeiger:
This adds the possibility to configure gradient clipping for the quantizers since it is a hyperparameter that changes the training.

~~Note: This PR is built on-top of #181.~~ rebased onto master and ready to be merged

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -33,6 +33,19 @@ We highly recommend using the first of these formulations: for the
 other two formulations, intermediate layers - like batch normalization or
 average pooling - and shortcut connections may result in non-binary input
 to the convolutions.
+
+Quantizers can eighter be referenced by string or called directly.
+The following usages are equivalent:
+
+```python
+lq.layers.QuantDense(64, kernel_quantizer="ste_sign")
+```
+```python
+lq.layers.QuantDense(64, kernel_quantizer=lq.quantizers.ste_sign)
+```
+```python
+lq.layers.QuantDense(64, kernel_quantizer=lq.quantizers.SteSign(clip_value=1.0))
+```
 """
 
 import tensorflow as tf
@@ -44,7 +57,15 @@ except:
     # For compatibility with TensorFlow 1.13
     from tensorflow.python.framework.tensor_util import is_tensor
 
-__all__ = ["ste_sign", "approx_sign", "magnitude_aware_sign", "SteTern", "ste_tern"]
+__all__ = [
+    "SteSign",
+    "ste_sign",
+    "approx_sign",
+    "MagnitudeAwareSign",
+    "magnitude_aware_sign",
+    "SteTern",
+    "ste_tern",
+]
 
 
 @tf.custom_gradient
@@ -95,9 +116,9 @@ class QuantizerFunctionWrapper:
 
 @utils.register_keras_custom_object
 @utils.set_precision(1)
-def ste_sign(x):
-    r"""
-    Sign binarization function.
+def ste_sign(x, clip_value=1.0):
+    r"""Sign binarization function.
+
     \\[
     q(x) = \begin{cases}
       -1 & x < 0 \\\
@@ -109,8 +130,8 @@ def ste_sign(x):
     (essentially the binarization is replaced by a clipped identity on the
     backward pass).
     \\[\frac{\partial q(x)}{\partial x} = \begin{cases}
-      1 & \left|x\right| \leq 1 \\\
-      0 & \left|x\right| > 1
+      1 & \left|x\right| \leq t_\text{clip} \\\
+      0 & \left|x\right| > t_\text{clip}
     \end{cases}\\]
 
     ```plot-activation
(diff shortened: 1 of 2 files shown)


larq/quantizers.py
```diff
@@ -33,19 +33,39 @@
 other two formulations, intermediate layers - like batch normalization or
 average pooling - and shortcut connections may result in non-binary input
 to the convolutions.
+
+Quantizers can eighter be referenced by string or called directly.
```
jamescook106: eighter -> either

larq/quantizers.py
```diff
@@ -72,8 +123,8 @@ def ste_sign(x):
     (essentially the binarization is replaced by a clipped identity on the
     backward pass).
     \\[\frac{\partial q(x)}{\partial x} = \begin{cases}
-      1 & \left|x\right| \leq 1 \\\
-      0 & \left|x\right| > 1
+      1 & \left|x\right| \leq t_\text{clip} \\\
```
jamescook106: Should be clip_value?

larq/quantizers.py
```diff
@@ -72,8 +123,8 @@ def ste_sign(x):
...
+      0 & \left|x\right| > t_\text{clip}
```
jamescook106: Should be clip_value?

larq/quantizers.py
```diff
@@ -82,6 +133,7 @@ def ste_sign(x):
 
     # Arguments
     x: Input tensor.
+    clip_value: Threshold for clipping gradients ($t_\text{clip}$).
```
jamescook106: Should be clip_value?

larq/quantizers.py
```diff
@@ -91,14 +143,49 @@ def ste_sign(x):
...
+      1 & \left|x\right| \leq t_\text{clip} \\\
+      0 & \left|x\right| > t_\text{clip}
+    \end{cases}\\]
+
+    ```plot-activation
+    quantizers.ste_sign
+    ```
+
+    # Arguments
+    clip_value: Threshold for clipping gradients ($t_\text{clip}$).
```
jamescook106: Should be clip_value?

larq/quantizers.py
```diff
@@ -108,6 +195,7 @@ def magnitude_aware_sign(x):
 
     # Arguments
     x: Input tensor
+    clip_value: Threshold for clipping gradients ($t_\text{clip}$).
```
jamescook106: Should be clip_value?

larq/quantizers.py
```diff
@@ -189,66 +298,92 @@ class SteTern:
     (essentially the Ternarization is replaced by a clipped identity on the
     backward pass).
     \\[\frac{\partial q(x)}{\partial x} = \begin{cases}
-    1 & \left|x\right| \leq 1 \\\
-    0 & \left|x\right| > 1
+    1 & \left|x\right| \leq t_\text{clip} \\\
+    0 & \left|x\right| > t_\text{clip}
     \end{cases}\\]
```
jamescook106: Should be clip_value?

larq/quantizers.py
lgeiger: I used clip_value here before and it looked not very good in the rendered version. I could live with a monospaced version that shows that it relates to the variable used in the code:
Bildschirmfoto 2019-08-14 um 14 16 44
I still prefer the t_clip since it is documented in the line below, what do you think?

jamescook106: I would always rather be explicit and use clip_value to reduce confusion even though aesthetically it looks worse

lgeiger:
> I mainly have questions, the big thing though is that I'm worried that moving to this has increased the complexity of the code, which is a barrier to use, especially since people may be coming from a non-python framework.

I agree that with this the code becomes a harder to maintain, although most of the changes are docstrings. But I don't think it will be a barrier to use, in fact the additional code mostly comes from making it easier to use (callable via string, function or class) in order to adhere with the Keras API conventions.

What I dislike about this PR though, is the need for two different APIs to do very similar things.
Unfortunately if we want to support configurable quantizers that can be serialized (e.g. when saving a model to a `.h5` file) we need to use classes. Keras does this for Constraints, Regularizers and Losses. For quantizers with trainable weights we will even need to subclass Layers similar to how Keras advanced activations work.
That means if we want to be inline with how Keras treats the activations, that are simple non-configurable functions that can be called by strings, I am afraid that we need the second function based API that then can also be used as a string. After writing this I am even considering removing this possibility, though this might not be a welcome change for most users 🤔

MariaHeuss:
I am not sure if we want to remove the possibility to call the functions as strings. I think this makes it easy to use especially when exploring Larq, when you are just experimenting around without knowing much about it. 
I do like that you can adjust the clip value now. 
I also do not think that this will make it harder to use.

lgeiger:
@MariaHeuss Thanks for the feedback. If you agree, it would be good to get this merged soon since a bunch of PRs are blocked by this.

PR #193, opened 2019-08-30
Format numbers in larq model summaries to be more readable. Closes #192.

### LINKED ISSUE #192: Human readable large numbers in `models.summary()`
At the moment the summary table generated by `models.summary()` has lines such as: `Number of MACs: 13228896000` where the raw integer is printed. It would be a lot more readable if this number were instead printed as `13.229 B` or `13.229 Billion`, or at least if there was a thousands separator (space or comma).

AdamHillier:
An example approach that formats numbers with three-significant figures and a magnitude (e.g. `k`, `M`, `B`, et cetera). This works for all numbers up to ~999 trillion. For larger numbers, scientific notation is used.

Illustrative examples:

```
>>> for i in range(1, 20):
...     num = int("1234567890123456789"[:i])
...     print("Number:", num, " " * (20 - i), "Formatted:", _number_as_readable_str(num))
...
Number: 1                     Formatted: 1
Number: 12                    Formatted: 12
Number: 123                   Formatted: 123
Number: 1234                  Formatted: 1.23 k
Number: 12345                 Formatted: 12.3 k
Number: 123456                Formatted: 123 k
Number: 1234567               Formatted: 1.23 M
Number: 12345678              Formatted: 12.3 M
Number: 123456789             Formatted: 123 M
Number: 1234567890            Formatted: 1.23 B
Number: 12345678901           Formatted: 12.3 B
Number: 123456789012          Formatted: 123 B
Number: 1234567890123         Formatted: 1.23 T
Number: 12345678901234        Formatted: 12.3 T
Number: 123456789012345       Formatted: 123 T
Number: 1234567890123456      Formatted: 1.23E+15
Number: 12345678901234567     Formatted: 1.23E+16
Number: 123456789012345678    Formatted: 1.23E+17
Number: 1234567890123456789   Formatted: 1.23E+18

>>> for i in range(1, 20):
...     num = int("99999999999999999999"[:i])
...     print("Number:", num, " " * (20 - i), "Formatted:", _number_as_readable_str(num))
...
Number: 9                     Formatted: 9
Number: 99                    Formatted: 99
Number: 999                   Formatted: 999
Number: 9999                  Formatted: 10.0 k
Number: 99999                 Formatted: 100 k
Number: 999999                Formatted: 1.00 M
Number: 9999999               Formatted: 10.0 M
Number: 99999999              Formatted: 100 M
Number: 999999999             Formatted: 1.00 B
Number: 9999999999            Formatted: 10.0 B
Number: 99999999999           Formatted: 100 B
Number: 999999999999          Formatted: 1.00 T
Number: 9999999999999         Formatted: 10.0 T
Number: 99999999999999        Formatted: 100 T
Number: 999999999999999       Formatted: 1.00E+15
Number: 9999999999999999      Formatted: 1.00E+16
Number: 99999999999999999     Formatted: 1.00E+17
Number: 999999999999999999    Formatted: 1.00E+18
Number: 9999999999999999999   Formatted: 1.00E+19
```

### DIFF
--- a/larq/models.py
+++ b/larq/models.py
@@ -51,7 +51,7 @@ def _bitsize_as_str(bitsize):
 def _number_as_readable_str(num):
     # The initial rounding here is necessary so that e.g. `999000` gets
     # formatted as `1.000 M` rather than `1000 k`
-    num = float("{:.3g}".format(num))
+    num = float(f"{num:.3g}")
 
     # For numbers less than 1000, output them directly, stripping any trailing
     # zeros and decimal places.
@@ -61,7 +61,7 @@ def _number_as_readable_str(num):
     # For numbers that are at least 1000 trillion (1 quadrillion) format with
     # scientific notation (3 s.f. = 2 d.p. in scientific notation).
     if num >= 1e15:
-        return "{:#.2E}".format(num)
+        return f"{num:#.2E}"
 
     # Count the magnitude.
     magnitude = 0
@@ -71,7 +71,7 @@ def _number_as_readable_str(num):
 
     # ':#.3g' formats the number with 3 significant figures, without stripping
     # trailing zeros.
-    num = "{:#.3g}".format(num).rstrip(".")
+    num = f"{num:#.3g}".rstrip(".")
     unit = ["", " k", " M", " B", " T"][magnitude]
     return num + unit
 
--- a/larq/models.py
+++ b/larq/models.py
@@ -48,6 +48,31 @@ def _bitsize_as_str(bitsize):
         raise NotImplementedError()
 
 
+def _number_as_readable_str(num):
+    # The initial rounding here is necessary so that e.g. `999000` gets
+    # formatted as `1.000 M` rather than `1000 k`
+    num = float("{:.4g}".format(num))
+
+    # For numbers that are at least 1000 trillion (1 quadrillion) format with
+    # scientific notation.
+    if num >= 1e15:
+        return "{:#.3E}".format(num)
+
+    # Count the magnitude.
+    magnitude = 0
+    while abs(num) >= 1000 and magnitude < 4:
+        magnitude += 1
+        num /= 1000.0
+
+    # ':#.4g' formats the number with 4 significant figures, without stripping
+    # trailing zeros.
+    num = "{:#.4g}".format(num)
+    unit = ["", " k", " M", " B", " T"][magnitude]
+
+    # The `rstrip`s ensure that integers < 1000 get formatted as-is.
+    return (num + unit).rstrip("0").rstrip(".")
+
+
 def _format_table_entry(x, units=1):
     try:
         assert not np.isnan(x)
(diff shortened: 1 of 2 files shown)


lgeiger:
That definitely looks a lot better :+1:

Though in some cases I think it still might be useful to know the exact number of parameters (e.g. when reimplementing a model from scratch as a sanity check).
What do you think about just using spaces to separate the numbers for better readability e.g `999 999 999`?
@koenhelwegen Do you have an opinion on that?

koenhelwegen:
Nice! I think this will almost always be preferable, but we could add a flag in case someone requires the precise numbers?

lgeiger:
> I think this will almost always be preferable, but we could add a flag in case someone requires the precise numbers?

Thinking about this again, I agree. I don't really want to introduce another flag anyway, and if we really need precise numbers we can still use the built in Keras model summary.

AdamHillier:
I've updated it to use three significant figures and added a flag so that the precise values can be used instead if wanted, but by default the readable totals are used. I added an extra test to cover the non-default -- the snapshot shows no difference with the previous behaviour when readable totals are disabled.

AdamHillier:
And I've edited my original comment to update the example formatted numbers.

AdamHillier:
After discussion with @lgeiger I've removed the flag - if necessary we can add it back in at a later date.

larq/models.py
```diff
@@ -48,6 +48,34 @@ def _bitsize_as_str(bitsize):
         raise NotImplementedError()
 
 
+def _number_as_readable_str(num):
+    # The initial rounding here is necessary so that e.g. `999000` gets
+    # formatted as `1.000 M` rather than `1000 k`
+    num = float("{:.3g}".format(num))
```
lgeiger:
```suggestion
    num = float(f"{num:.3g}")
```

larq/models.py
```diff
@@ -48,6 +48,34 @@ def _bitsize_as_str(bitsize):
...
+
+    # For numbers less than 1000, output them directly, stripping any trailing
+    # zeros and decimal places.
+    if num < 1000:
+        return str(num).rstrip("0").rstrip(".")
+
+    # For numbers that are at least 1000 trillion (1 quadrillion) format with
+    # scientific notation (3 s.f. = 2 d.p. in scientific notation).
+    if num >= 1e15:
+        return "{:#.2E}".format(num)
```
lgeiger:
```suggestion
        return f"{num:#.2E}"
```

larq/models.py
```diff
@@ -48,6 +48,34 @@ def _bitsize_as_str(bitsize):
...
+
+    # Count the magnitude.
+    magnitude = 0
+    while abs(num) >= 1000 and magnitude < 4:
+        magnitude += 1
+        num /= 1000.0
+
+    # ':#.3g' formats the number with 3 significant figures, without stripping
+    # trailing zeros.
+    num = "{:#.3g}".format(num).rstrip(".")
```
lgeiger:
```suggestion
    num = f"{num:#.3g}".rstrip(".")
```

AdamHillier:
> What do you think about using f-strings instead of the .format syntax?

Yep that looks nicer, I've updated.

PR #195, opened 2019-08-30
Test against TensorFlow 2.0.0-rc0

lgeiger:


lgeiger:
This failure looks like it is related to to our training metrics. Though I am not sure if this is a problem with our code or with TF 2rc0.

lgeiger:
I have a workaround and posted an issue on TensorFlow to make sure this behaviour change is intended and not a bug: https://github.com/tensorflow/tensorflow/issues/32210

PR #203, opened 2019-09-02
Remove birealnet

koenhelwegen:


docs/models/examples.md
```diff
@@ -1,15 +1,17 @@
...
 ```python
 import tensorflow as tf
 import numpy as np
 import larq_zoo as lqz
 
-model = lqz.BiRealNet(weights="imagenet")
+model = lqz.BinaryDenseNet45(weights="imagenet")
 
-img_path = "tests/fixtures/elephant.jpg"
+img_path = "path/to/elephant.jpg"
```
lgeiger: If the example is run from the repo directory, it will find the image. But I'd also be OK to change this.

docs/models/examples.md
```diff
@@ -61,15 +63,15 @@ x = np.expand_dims(x, axis=0)
 average_pool_8_features = model.predict(x)
 ```
 
-## Fine-tune Bi-Real Net on a new set of classes
+## Fine-tune BinaryDenseNet-45 on a new set of classes
```
lgeiger: Have you checked, that the layer numbers in this example make sense with the new model?

docs/models/index.md
```diff
@@ -18,7 +18,6 @@ The model definitions and the train loop are available in the [Larq Zoo reposito
 | [BinaryDenseNet37](/api/larq_zoo/#binarydensenet37)               | 62.89 %        | 84.19 %        | 8 734 120  | 5.25 MB  |
 | [BinaryDenseNet28](/api/larq_zoo/#binarydensenet28)               | 60.91 %        | 82.83 %        | 5 150 504  | 4.12 MB  |
 | [BinaryResNetE18](/api/larq_zoo/#binaryresnete18)                 | 58.32 %        | 80.79 %        | 11 699 368 | 4.03 MB  |
-| [Bi-Real Net](/api/larq_zoo/#birealnet)                           | 58.01 %        | 80.27 %        | 11 736 232 | 4.04 MB  |
```
lgeiger: I don't think we should remove it from here.

docs/models/examples.md
koenhelwegen: Would have to be the `larq-zoo` directory even though this documentation now lives in `larq`, and if people read this from the website I would also say this is clearer.

lgeiger: OK, makes sense

koenhelwegen:
No longer needed due to #213

PR #205, opened 2019-09-03
Unify colour scheme and fix some minor styling issues

lgeiger:
This PR properly configures the colour theme and fixes some minor rendering issues on mobile devices.

The CSS and HTML is mostly copy past from the theme with minor modifications related to the colour and the logo dimensions

### DIFF
--- a/docs/custom.css
+++ b/docs/custom.css
@@ -8,7 +8,102 @@
   transition: unset;
   box-shadow: 0 0 0.2rem rgba(0, 0, 0, 0.1), 0 0.2rem 0.4rem rgba(0, 0, 0, 0.2);
 }
-
-[data-md-color-primary=grey] .md-header, [data-md-color-primary=grey] .md-hero {
+button[data-md-color-primary="larq"] {
   background-color: #0a1d3c;
+}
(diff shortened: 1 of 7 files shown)


docs/custom.css
```diff
@@ -8,7 +8,102 @@
...
   background-color: #0a1d3c;
 }
+[data-md-color-accent="larq"]
+  .md-sidebar__scrollwrap::-webkit-scrollbar-thumb:hover {
+  background-color: #059e29;
+}
+[data-md-color-accent="larq"] .md-source-file:hover:before {
+  background-color: #059e29;
+}
+@media only screen and (max-width: 59.9375em) {
```
AdamHillier: It seems weird to me to have `em`/`rem` dimensions which are so specific. Not an issue, just interesting! It looks like it's been calculated as 959px / 16pt font size . I guess doing it with `em` improves accessibility for non-default font-sizes.

lgeiger: I am also unsure about this. This is mostly copy paste from the default theme, so I don't know what the exact reasoning is behind this.

PR #209, opened 2019-09-06
Fix MAC count strings. Fixes #208.

### LINKED ISSUE #208: MAC output strings are wrong
In the table, but not in the summary, we are incorrectly dividing by `8 * 1024` and displaying a count in 'kB'.

AdamHillier:


AdamHillier:
If you want smaller numbers in the table rows I can apply the `_number_as_readable_str` method.

PR #210, opened 2019-09-06
Fix Tensor comparison to support TF2-rc0 hashing behaviour

lgeiger:
Recently TensorFlow changed the behaviour of equality in 2.0 such that Variables & Tensors are no longer hashable. This breaks our current strategy of Tensor comparison, since TF 2 now tries to compare elementwise. This PR fixes the comparison to account for the new behaviour.

For more info see https://github.com/tensorflow/tensorflow/issues/32210
Closes #195

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,7 +19,7 @@ jobs:
           pytest.args: "--cov=larq --cov-report=html --cov-config=.coveragerc"
         Python37_TF2:
           python.version: "3.7"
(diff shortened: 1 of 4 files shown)


lgeiger:
@koenhelwegen Could you check what's going on with the model summary here?

koenhelwegen:
For some reason, `layer.quantized_latent_weights().index(kernel)` works for any kernel except the pointwise kernel in the separable convolution.. Searching for the name of the weight instead solves the problem.

PR #211, opened 2019-09-07
Replace weight list search by weight name list search

koenhelwegen:


lgeiger:
Thanks a lot for debugging this 👍 

I am not sure wether using comparison by name works in a distributed context. What do you think about the following fix: 6f863a84712082e1f015ccbd099b315855b3e915 ?

koenhelwegen:
Good point, this is better

PR #212, opened 2019-09-08
Simplify string formatting

lgeiger:
As far as I know, this shouldn't change functionallity, but removing the `#` fixes syntax highlighting inside my text editor.

### DIFF
--- a/larq/models.py
+++ b/larq/models.py
@@ -61,7 +61,7 @@ def _number_as_readable_str(num):
     # For numbers that are at least 1000 trillion (1 quadrillion) format with
     # scientific notation (3 s.f. = 2 d.p. in scientific notation).
     if num >= 1e15:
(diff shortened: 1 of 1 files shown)


AdamHillier:
Hmm so I think this does change functionality. The `#` prevents the formatter from stripping trailing zeroes (and potentially the decimal place), so with your changes I get: 

```
>>> _number_as_readable_str(1000)
'1 k'
```
but with the original I get:
```
>>> _number_as_readable_str(1000)
'1.00 k'
```

It's not terrible, but I prefer the original because it's consistently three significant figures. Of course, this is only ever a problem if every digit from the third onwards is zero, which is pretty unlikely for sufficiently large uniformly random numbers.

What do you think?

lgeiger:
Makes sense, let's close it 👍

PR #216, opened 2019-09-10
Add TensorFlow 1.15.0-rc0 to test matrix

lgeiger:


azure-pipelines.yml
```diff
@@ -12,14 +12,17 @@ jobs:
...
-          tensorflow.version: "1.13.1"
+          tensorflow: "tensorflow==1.13.1"
         Python37_TF114:
           python.version: "3.7"
-          tensorflow.version: "1.14.0"
+          tensorflow: "tensorflow==1.14.0"
           pytest.args: "--cov=larq --cov-report=html --cov-config=.coveragerc"
+        Python37_TF115:
+          python.version: "3.7"
+          tensorflow: "tensorflow-cpu==1.15.0-rc0"
```
lgeiger: By default `pip install tensorflow` now includes GPU support and doesn't fail on machines without NVIDIA GPUs. The `tensorflow-cpu` package is smaller in size, so I guess it will be faster to install on CI.

PR #223, opened 2019-09-16
Add Dorefa quantizer

MariaHeuss:
This adds the k_bit quantizer from the DoReFa paper. https://arxiv.org/abs/1606.06160
The _dorefa_quantizer is a private function since I did not find an elegant way to set the precision dependent on the k_bit argument, so I think this function always be initialized with the DorefaQuantizer class where the precision is set dependent on this argument.

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -69,6 +69,7 @@ __all__ = [
     "ste_tern",
     "SteHeaviside",
     "ste_heaviside",
+    "DorefaQuantizer"
 ]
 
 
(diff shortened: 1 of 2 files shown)


larq/quantizers.py
```diff
@@ -532,6 +533,60 @@ def grad(dy):
...
+
+    where \\(n = 2^{\text{k_bit}} - 1\\). The number of bits, k_bit, needs to be passed as an argument.
+    The gradient is estimated using the Straight-Through Estimator
+    (essentially the binarization is replaced by a clipped identity on the
+    backward pass).
+    \\[\frac{\partial q(x)}{\partial x} = \begin{cases}
+    1 &  0 \leq x \leq 1 \\\
+    0 & \text{else}
+    \end{cases}\\]
+    Note that in the original tensorpack implementation the clip and
```
lgeiger: I'd remove that comment
```suggestion
```

larq/quantizers.py
```diff
@@ -532,6 +533,60 @@ def grad(dy):
...
+    the quantizer are two different functions that get composed.
```
```suggestion
```

larq/quantizers_test.py
```diff
@@ -202,3 +202,38 @@ def test_magnitude_aware_sign():
...
+)
+def test_dorefa_quantize(fn):
+    x = tf.keras.backend.placeholder(ndim=2)
+    f = tf.keras.backend.function([x], [fn(x)])
+    real_values = np.random.uniform(-2, 2, (2, 5))
+    result = f([real_values])[0]
+    k_bit = 2
+    n = 2 ** k_bit - 1
+    assert not np.any(result > 1)
+    assert not np.any(result < -1)
```
lgeiger: Shouldn't this be:
```suggestion
    assert not np.any(result < 0)
```

MariaHeuss: No, actually this quantizer only outputs values in the [0,1] interval. It sounds a bit odd but works... s

MariaHeuss:
> I think it is fine to make the `do_re_fa_quantizer` public and just set the precision from the default argument. If the function is used with a keras model there is no way to change the precision anyway so the precision of the default argument will be correctly displayed. If it is used outside of a keras model the summary won't work anyway.

I do not really get the argument with the keras model but I am fine with making it a public function because people will realize latest when they see the model summary that something went wrong when were using the function instead of the class.

larq/quantizers.py
MariaHeuss: I guess you suggest to delete the whole sentence right?

MariaHeuss:
> I think it is fine to make the `do_re_fa_quantizer`

Do you prefer the name do_re_ra_quantizer over dorefa_quantizer?

larq/quantizers.py
```diff
@@ -536,6 +538,91 @@ def grad(dy):
     return _and_binarize_with_identity_grad(x)
 
 
+def dorefa_quantizer(x, k_bit=2):
```
```suggestion
@utils.register_keras_custom_object
@utils.set_precision(2)
def dorefa_quantizer(x, k_bit=2):
```

MariaHeuss:
I accidentally commited the change without real message. I will leave it like this since we are going to squash merge it anyways.

larq/quantizers_test.py
lgeiger: > No, actually this quantizer only outputs values in the [0,1] interval.

Yes, but the above test only checks that there are no values smaller than -1 instead of it should assert that we don't have any values smaller 0. Or am I missing something?

lgeiger:
> Do you prefer the name do_re_ra_quantizer over dorefa_quantizer? I prefer keeping dorefa_quantizer but this is just a weak preference.

I agree. `dorefa_quantizer` is nicer even if it doesn't 100% follow PEP8.

larq/quantizers_test.py
MariaHeuss: No, I just was confused about the colors I guess. You're right of course.

PR #225, opened 2019-09-17
Created CompatibilityMetric class, added register_alias decorator

jneeven:
CompatibilityMetric will be the parent class of all our custom metrics. I added a register_alias decorator and registered the FlipRatio under flip_ratio, such that we can find it with tf.keras.metrics.get() in the future.

### DIFF
--- a/larq/metrics.py
+++ b/larq/metrics.py
@@ -62,7 +62,47 @@ def get_training_metrics():
     return _GLOBAL_TRAINING_METRICS
 
 
-class FlipRatio(Metric):
+class _CompatibilityMetric(Metric):
+    r"""Metric with support for both 1.13 and 1.14+
+    """
+
+    def add_weight(
+        self,
+        name,
+        shape=(),
+        aggregation=tf.VariableAggregation.NONE,
+        synchronization=tf.VariableSynchronization.ON_READ,
+        initializer=None,
+        dtype=None,
+    ):
+        if utils.tf_1_14_or_newer():
+            return super().add_weight(
+                name=name,
+                shape=shape,
+                aggregation=aggregation,
+                synchronization=synchronization,
+                initializer=initializer,
+                dtype=dtype,
+            )
+        else:
+            # Call explicitely tf.keras.layers.Layer.add_weight because TF 1.13
+            # doesn't support setting a custom dtype
+            return tf.keras.layers.Layer.add_weight(
+                self,
+                name=name,
+                shape=shape,
+                dtype=self._dtype if dtype is None else dtype,
+                trainable=False,
+                initializer=initializer,
+                collections=[],
+                synchronization=synchronization,
+                aggregation=aggregation,
+            )
+
+
+@utils.register_alias("flip_ratio")
+@utils.register_keras_custom_object
+class FlipRatio(_CompatibilityMetric):
     """Computes the mean ration of changed values in a given tensor.
 
     !!! example
@@ -93,12 +133,17 @@ class FlipRatio(Metric):
                 shape=values_shape,
                 dtype=self.values_dtype,
                 initializer=tf.keras.initializers.zeros,
+                aggregation=tf.VariableAggregation.SUM,
             )
             self.total = self.add_weight(
-                "total", initializer=tf.keras.initializers.zeros
+                "total",
+                initializer=tf.keras.initializers.zeros,
+                aggregation=tf.VariableAggregation.SUM,
             )
             self.count = self.add_weight(
-                "count", initializer=tf.keras.initializers.zeros
+                "count",
+                initializer=tf.keras.initializers.zeros,
+                aggregation=tf.VariableAggregation.SUM,
             )
         self._size = np.prod(self.values_shape)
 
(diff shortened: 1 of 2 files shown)


larq/metrics.py
```diff
@@ -62,7 +63,54 @@ def get_training_metrics():
...
+class _CompatibilityMetric(Metric):
+    r"""Metric with support for both 1.13 and 1.14+
+    """
+
+    def get_config(self):
+        """Returns the serializable config of the metric."""
+        return {**super().get_config(), "name": self.name, "dtype": self.dtype}
+
+    def reset_states(self):
+        return K.batch_set_value([(v, 0) for v in self.variables])
```
lgeiger: Since we only use it here, I'd prefer:
```suggestion
        return tf.keras.backend.batch_set_value([(v, 0) for v in self.variables])
```
But that's just personal preference.

larq/metrics.py
```diff
@@ -62,7 +63,54 @@ def get_training_metrics():
     return _GLOBAL_TRAINING_METRICS
 
 
-class FlipRatio(Metric):
+class _CompatibilityMetric(Metric):
```
lgeiger: Thinking about this again what do you think about `LarqMetric`?

larq/metrics.py
```diff
@@ -62,7 +63,54 @@ def get_training_metrics():
...
+    r"""Metric with support for both 1.13 and 1.14+
+    """
```
lgeiger:
```suggestion
```

larq/metrics.py
```diff
@@ -62,7 +63,54 @@ def get_training_metrics():
     return _GLOBAL_TRAINING_METRICS
 
 
-class FlipRatio(Metric):
+class _CompatibilityMetric(Metric):
+    r"""Metric with support for both 1.13 and 1.14+
```
lgeiger:
```suggestion
    """Metric with support for both 1.13 and 1.14+"""
```

larq/metrics.py
```diff
@@ -62,7 +63,54 @@ def get_training_metrics():
...
+    """
+
+    def get_config(self):
```
lgeiger: I don't think we need to overwrite this since this is the default anyway:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/metrics.py#L199-L201

larq/metrics.py
```diff
@@ -62,7 +63,54 @@ def get_training_metrics():
...
+                dtype=self._dtype if dtype is None else dtype,
+                trainable=False,
+                initializer=initializer,
+                collections=[],
+                synchronization=synchronization,
+                aggregation=aggregation,
+            )
+
+
+@utils.register_alias("flip_ratio")
```
lgeiger: Nice!

larq/metrics.py
```diff
@@ -87,18 +135,24 @@ def __init__(
         super().__init__(name=name, dtype=dtype)
         self.values_dtype = tf.as_dtype(values_dtype)
         self.values_shape = tf.TensorShape(values_shape).as_list()
+        self.is_weight_metric = True
```
lgeiger: This isn't used anywhere. Or am I missing something?
```suggestion
```

larq/metrics.py
```diff
@@ -87,18 +135,24 @@ def __init__(
...
         with tf.init_scope():
             self._previous_values = self.add_weight(
                 "previous_values",
                 shape=values_shape,
                 dtype=self.values_dtype,
                 initializer=tf.keras.initializers.zeros,
+                aggregation=tf.VariableAggregation.SUM,
```
lgeiger: I am not sure if this is correct. Could you elaborate on this change?

larq/metrics.py
jneeven: Not in this PR, but I need it for the MetricLayer. Didn't want to push them all at the same time since the changes might be a bit overwhelming then

larq/metrics.py
jneeven: You already had it on SUM before, it was the default value in the add_weight function. However, SUM may not always be the desired aggregation method, so I changed the default to None (which will throw an error if you run it on multi-gpu, rather than computing incorrect values) and explicitly set it back to SUM here. 
That being said, I think it is actually incorrect here; I don't think FlipRatio can support multi-GPU at all. In that case, I should set it to None. What do you think?

larq/metrics.py
jneeven: Sure, at this point there is nothing that really makes it larq-specific, but since all Larq metrics will subclass it, why not

larq/metrics.py
jneeven: Alternatively, we could set all of these to ONLY_FIRST_REPLICA, which will use only one of the results in the multi-gpu case and ignore the others.

jneeven: I changed it to ONLY_FIRST_REPLICA in the latest commit. Since the FlipRatio uses weights instead of activations, I assume that even across multiple GPU's, either the metric would only be computed once, or all replicas would compute the same value.

larq/metrics.py
```diff
@@ -62,7 +62,49 @@ def get_training_metrics():
...
+    """Metric with support for both 1.13 and 1.14+"""
+
+    def reset_states(self):
+        return tf.keras.backend.batch_set_value([(v, 0) for v in self.variables])
+
+    def add_weight(
+        self,
+        name,
+        shape=(),
+        aggregation=tf.VariableAggregation.NONE,
```
lgeiger: I'd prefer not changing the default here, so that `LarqMetric` will behave the same as `keras.metrics.Metric`.
```suggestion
        aggregation=tf.VariableAggregation.SUM,
```

larq/metrics.py
lgeiger: I think switching to `ONLY_FIRST_REPLICA` makes sense here :+1:

larq/metrics.py
```diff
@@ -62,7 +62,49 @@ def get_training_metrics():
     return _GLOBAL_TRAINING_METRICS
 
 
-class FlipRatio(Metric):
+class LarqMetric(Metric):
+    """Metric with support for both 1.13 and 1.14+"""
+
+    def reset_states(self):
```
lgeiger: We can also remove this method since it is implemented in the parent class already:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/metrics.py#L203-L209

jneeven: Agreed, had not noticed this

PR #231, opened 2019-09-19
Use Altair/Vega to plot activations and quantizers in API docs

lgeiger:
This switches the activation and quantizer plots in the API docs from matplotlib to Altair/Vega to make them consistent with the rest of the plots in our docs.

This will also allow us in the future to add interactive tooltips to the plot if we need to.


lgeiger:
Here are the links to compare:
- This PR: https://larq-i9q6bt5op.now.sh/api/activations/#leaky_tanh
- Master: https://larq.dev/api/activations/#leaky_tanh

PR #232, opened 2019-09-19
Add option to use unclipped identity STE in quantizers

MariaHeuss:
This will add the option for using a non clipped STE in all quantizers where it makes sense.

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -146,7 +146,7 @@ def ste_sign(x, clip_value=1.0):
 
     # Arguments
     x: Input tensor.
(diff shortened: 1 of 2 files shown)


lgeiger:
Could you also adapt the docstrings to mention what `clip_value=None` does. Maybe something like:
```
clip_value: Threshold for clipping gradients. If `None` gradients are not clipped.
```

MariaHeuss:
> Could you also adapt the docstrings to mention what `clip_value=None` does. Maybe something like:

Sure, will do!

PR #239, opened 2019-09-20
Change "Why Larq?" menu item to "About"

lgeiger:
Having a FAQ under "Why Larq?" is not really intuitive. I think changing the heading to "About" makes sense or should we name it differently?

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -32,7 +32,7 @@ nav:
       - Papers using Larq: papers.md
       - Contributing Guide: contributing.md
       - Code of Conduct: code-of-conduct.md
-  - Why Larq?:
+  - About:
       - About: about.md
       - FAQ: faq.md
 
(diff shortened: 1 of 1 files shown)


lgeiger:
> The sidebar in the about page. now looks a little weird. Not sure how to fix it though.

I agree. We could change the title of the about page to something more descriptive, but I am also not sure what the best title would be. Any suggestions?

jneeven:
I'd suggest just switching them around: the title becomes "About", and the title of what is currently "about" (https://larq.dev/about/) can change to "Why Larq?"

lgeiger:
> I'd suggest just switching them around: the title becomes "About", and the title of what is currently "about" (https://larq.dev/about/) can change to "Why Larq?"

I changed it. @AdamHillier What do you think?

AdamHillier:
Yep looks better

PR #245, opened 2019-09-20
Added GradientFlow metric

jneeven:
For some reason, the changed files include the code of conduct. Not sure what's going on there...


lgeiger:
> For some reason, the changed files include the code of conduct. Not sure what's going on there...

Looks like `git` committed the files and did not correctly recognise the symlinks. Could you rebase and revert the changes to the markdown files?

larq/metrics.py
```diff
@@ -172,3 +172,60 @@ def get_config(self):
...
+        name: Name of the metric.
+    """
+
+    def __init__(self, threshold=1.0, name="gradient_flow", dtype=None):
+        super().__init__(name=name, dtype=dtype)
+
+        self.threshold = float(threshold)
+
+        with tf.init_scope():
+            self._threshold = tf.constant(self.threshold, dtype=self.dtype)
```
lgeiger: Why do we need to cast this to a Tensor here?
I think TensorFlow will already handle the casting automatically in line 216

larq/metrics.py
```diff
@@ -172,3 +172,60 @@ def get_config(self):
             "values_shape": self.values_shape,
             "values_dtype": self.values_dtype.name,
         }
+
+
+@utils.register_alias("gradient_flow")
+@utils.register_keras_custom_object
+class GradientFlow(LarqMetric):
+    r"""Indicator of gradient mismatch and saturation as described in https://arxiv.org/pdf/1904.02823.pdf.
+        Counts the ratio of activations for each neuron (i.e. pixel in a feature map) over the entire batch
```
lgeiger: I'd prefer to remove one level of indentation here. It would be great if `black` could also handle this.

larq/metrics.py
```diff
@@ -172,3 +172,60 @@ def get_config(self):
...
+
+        self.threshold = float(threshold)
+
+        with tf.init_scope():
+            self._threshold = tf.constant(self.threshold, dtype=self.dtype)
+            self.total_value = self.add_weight(
+                "total_value",
+                initializer=tf.keras.initializers.zeros,
+                dtype=self.dtype,
+                aggregation=tf.VariableAggregation.SUM,
```
lgeiger: Since this is the default, we can remove this line.
```suggestion
```

larq/metrics.py
```diff
@@ -172,3 +172,60 @@ def get_config(self):
...
+            )
+            self.num_batches = self.add_weight(
+                "num_batches",
+                initializer=tf.keras.initializers.zeros,
+                dtype=tf.int32,
+                aggregation=tf.VariableAggregation.SUM,
```
lgeiger: See above
```suggestion
```

larq/metrics.py
```diff
@@ -172,3 +172,60 @@ def get_config(self):
...
+            )
+
+    def update_state(self, values):
+        values = tf.cast(values, self.dtype)
+
+        below_threshold = tf.math.count_nonzero(
+            tf.math.abs(values) <= self._threshold, dtype=tf.float32
```
lgeiger: Shouldn't this be the following?
```suggestion
            tf.math.abs(values) <= self._threshold, dtype=self.dtype
```
I didn't know that the casting here is possible. That's handy :+1:

larq/metrics.py
```diff
@@ -172,3 +172,60 @@ def get_config(self):
...
+        )
+        num_activations = tf.cast(tf.size(values), tf.float32)
```
lgeiger:
```suggestion
        num_activations = tf.cast(tf.size(values), self.dtype)
```

larq/metrics.py
```diff
@@ -172,3 +172,60 @@ def get_config(self):
...
+        ratio = tf.cast(below_threshold / num_activations, self.dtype)
```
lgeiger: If we can do the above change, we don't need to cast it here again.
```suggestion
        ratio = below_threshold / num_activations
```

larq/metrics.py
```diff
@@ -172,3 +172,60 @@ def get_config(self):
...
+            self.total_value = self.add_weight(
+                "total_value",
+                initializer=tf.keras.initializers.zeros,
+                dtype=self.dtype,
+                aggregation=tf.VariableAggregation.SUM,
+            )
+            self.num_batches = self.add_weight(
+                "num_batches",
+                initializer=tf.keras.initializers.zeros,
+                dtype=tf.int32,
```
lgeiger: I think it's fine to store it as float, then we don't need to cast it during run time.
That's how TensorFlow handles it internally: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/metrics.py#L297
```suggestion
```

larq/metrics.py
```diff
@@ -172,3 +172,60 @@ def get_config(self):
...
+        super().__init__(name=name, dtype=dtype)
+
+        self.threshold = float(threshold)
+
+        with tf.init_scope():
+            self._threshold = tf.constant(self.threshold, dtype=self.dtype)
+            self.total_value = self.add_weight(
+                "total_value",
+                initializer=tf.keras.initializers.zeros,
+                dtype=self.dtype,
```
lgeiger: This is the default anyway:
```suggestion
```

larq/metrics.py
```diff
@@ -172,3 +172,60 @@ def get_config(self):
...
+        num_activations = tf.cast(tf.size(values), tf.float32)
+        ratio = tf.cast(below_threshold / num_activations, self.dtype)
+
+        update_total_op = self.total_value.assign_add(ratio)
+        with tf.control_dependencies([update_total_op]):
+            return self.num_batches.assign_add(1)
+
+    def result(self):
+        return tf.compat.v1.div_no_nan(
+            tf.cast(self.total_value, tf.float32), tf.cast(self.num_batches, tf.float32)
```
lgeiger: If we store `num_batches` as a floating value we don't need to cast here.

larq/metrics.py
jneeven: I thought it might be better in terms of performance, since we're adding this integer every batch and only request the float at the end of the epoch.

lgeiger: Hmm, might be true. Though I'd stick to what TF is doing internally.

larq/metrics.py
jneeven: Should it? I always want to calculate the ratio in floats. self.dtype should only affect what the type of the final result should be, right? You could specify an integer as dtype, and then the calculations will be very wrong. (Although that would be pointless anyway since the value is always between 0 and 1)

larq/metrics.py
jneeven: @koenhelwegen  any opinions on this?

jneeven: I suppose the performance impact is negligible, since we're only adding 1 number per batch. I'll change it to float then

larq/metrics.py
lgeiger: I am unsure. I think this will always be either float32 or float16.

lgeiger:
Looks like this needs an other round of `black` formatting ;)

larq/metrics.py
```diff
@@ -172,3 +172,51 @@ def get_config(self):
...
+            )
+            self.num_batches = self.add_weight(
+                "num_batches", initializer=tf.keras.initializers.zeros
+            )
+
+    def update_state(self, values):
+        values = tf.cast(values, self.dtype)
+
+        below_threshold = tf.math.count_nonzero(
+            tf.math.abs(values) <= self.threshold, dtype=self.dtype
```
lgeiger: Do you think computing the value over the entire batch is the correct thing to do here?

jneeven: Yeah, we could compute the average per channel or per sample as well, but they eventually get averaged over all dimensions anyway (since the metric needs to be a scalar). The order in which we take the average doesn't matter, so we might as well compute it over the entire batch immediately.

larq/metrics.py
```diff
@@ -172,3 +172,51 @@ def get_config(self):
             "values_shape": self.values_shape,
             "values_dtype": self.values_dtype.name,
         }
+
+
+@utils.register_alias("gradient_flow")
+@utils.register_keras_custom_object
+class GradientFlow(LarqMetric):
```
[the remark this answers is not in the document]
jneeven: @lgeiger do you know if keras supports calling a metric on gradient values? I can't quite figure it out from the [add_metric function](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/engine/base_layer.py#L1226), but I assume that calling add_metric in the backward pass will break tensorflow...

larq/metrics.py
```diff
@@ -12,7 +12,7 @@
 __all__ = ["scope", "get_training_metrics"]
```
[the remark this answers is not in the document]
jneeven: It seems all custom metrics will currently break on multi-gpu with tf 1.14.0, so perhaps it's better not to document them until there is an official release for 1.15 or 1.2 (and until we have some infrastructure in place to automatically disable them for 1.14.0 + multi-gpu)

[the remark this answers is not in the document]
lgeiger: @jneeven Agreed

larq/metrics.py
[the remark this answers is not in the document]
lgeiger: @koenhelwegen I agree, that might be a better API, though I am also not sure how to implement this in Keras without using a custom training loop.

@jneeven I don't know. This might be good to ask in the TF/Keras mailing list.

In TensorFlow 1.x it was possible to get the gradients in the [TensorBoard callback](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/callbacks_v1.py#L211-L223). In TensorFlow 2.x [this options is disabled](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/callbacks.py#L1478-L1480).

lgeiger:
This has been open for a very long time now. I think #343 might be a solution for getting this merged.

lgeiger:
This has been open for quite a while now.
Since #402 is merged now, @jneeven do you think we should revisit this with the implementation mentioned in #343 which would resolve https://github.com/larq/larq/pull/245#discussion_r327550655?

jneeven:
> This has been open for quite a while now.
> Since #402 is merged now, @jneeven do you think we should revisit this with the implementation mentioned in #343 which would resolve [#245 (comment)](https://github.com/larq/larq/pull/245#discussion_r327550655)?

Yes, we definitely should. I won't have time for this the next few days, but feel free to take over this PR. Otherwise, I'll hopefully do it within now and two weeks :sweat_smile:

MariaHeuss:
I can take over from here.

jneeven:
After almost a year of inactivity, I'm closing this PR. If we do decide to add such a metric in the future, it will be in a much more end-to-end manner that actually looks at the gradients rather than computing them in-place using some hyperparameters.

PR #253, opened 2019-09-25
DoReFa-Net

MariaHeuss:
This adds the docs for the DoReFa-Net model.


jamescook106:
Is this 1 bit weights, 2 bit activations?

PR #267, opened 2019-10-02
Explicitely clip gradients of quantizers in backwards pass

lgeiger:
This removes the call to `tf.clip_by_value` in the forward pass and replaces it with explicit gradient clipping. This gives a minor speedup in the forward pass and optimizes the gradient computation.
~~I think this also makes the implementation a bit more readable.~~ Unfortunately `@tf.custom_gradient` only supports key word arguments in eager execution, that makes the code a bit more complicated.

The gradients are computed as done in [`tf.clip_by_value` which calls `tf.math.minimum` and `tf.math.maximum` under the hood](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/math_grad.py#L1417-L1473).

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -118,7 +118,6 @@ class QuantizerFunctionWrapper:
 
 @utils.register_keras_custom_object
 @utils.set_precision(1)
-@tf.custom_gradient
 def ste_sign(x, clip_value=1.0):
     r"""Sign binarization function.
 
@@ -153,10 +152,14 @@ def ste_sign(x, clip_value=1.0):
       Activations Constrained to +1 or -1](http://arxiv.org/abs/1602.02830)
     """
 
-    def grad(dy):
-        return _clipped_gradient(x, dy, clip_value)
+    @tf.custom_gradient
+    def _call(x):
+        def grad(dy):
+            return _clipped_gradient(x, dy, clip_value)
 
-    return math.sign(x), grad
+        return math.sign(x), grad
+
+    return _call(x)
 
 
 @utils.register_keras_custom_object
@@ -300,7 +303,6 @@ def approx_sign(x):
 
 @utils.register_keras_custom_object
 @utils.set_precision(1)
-@tf.custom_gradient
 def swish_sign(x, beta=5.0):
     r"""Sign binarization function.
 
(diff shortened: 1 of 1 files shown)


larq/quantizers.py
```diff
@@ -74,17 +74,13 @@
...
-def _binarize_with_weighted_grad(x):
-    def grad(dy):
-        return (1 - tf.abs(x)) * 2 * dy
+def _clipped_gradient(x, dy, clip_value):
+    if clip_value is None:
+        return dy
 
-    return math.sign(x), grad
+    zeros = tf.zeros_like(dy)
+    mask = tf.math.less_equal(tf.math.abs(x), clip_value)
```
lgeiger: This could also be done using
```python
mask = tf.math.logical_and(
        tf.math.less_equal(x, clip_value), tf.math.greater_equal(x, -clip_value)
    )
```
I roughly tested this on a GPU where usign `tf.abs` was faster. This solution is slightly faster on my non optimized mac CPU build.

larq/quantizers.py
```diff
@@ -74,17 +74,13 @@
...
+    return tf.where(mask, dy, zeros)
```
lgeiger: We could also explicitly call `where_v2` here, but I think that makes it a lot more readable.

larq/quantizers.py
```diff
@@ -156,10 +153,10 @@ def ste_sign(x, clip_value=1.0):
       Activations Constrained to +1 or -1](http://arxiv.org/abs/1602.02830)
     """
 
-    if clip_value is not None:
-        x = tf.clip_by_value(x, -clip_value, clip_value)
+    def grad(dy):
+        return _clipped_gradient(x, dy, clip_value)
 
-    return _binarize_with_identity_grad(x)
+    return math.sign(x), grad
```
lgeiger: We could also use a lambda function here, but I think a explicit `grad` function is more readable for people with less experience with `ﬂtf.custom_gradient`.

larq/quantizers.py
koenhelwegen: would keep the current implementation as it's a bit cleaner.

larq/quantizers.py
koenhelwegen: API is the same right? If `where_v2` is faster/recommended I would use that.

Tombana: I think `where` and `where_v2` are equally readable so lets go for whichever one is faster.

lgeiger: `tf.where == tf.where_v2 == tf.compat.v2.where` on TensorFlow 2.0.0. Unfortunately it is not available in TF 1.13 so it would be messy to use it. This comment was mostly for reference.

larq/quantizers.py
```diff
@@ -156,10 +152,14 @@ def ste_sign(x, clip_value=1.0):
...
-        x = tf.clip_by_value(x, -clip_value, clip_value)
+    @tf.custom_gradient
+    def _call(x):
+        def grad(dy):
+            return _clipped_gradient(x, dy, clip_value)
 
-    return _binarize_with_identity_grad(x)
+        return math.sign(x), grad
+
+    return _call(x)
```
Tombana: Is there a reason that `ste_sign` is implemented as `return _call(x)` where `_call` gets the `tf.custom_gradient` decorator, whereas `approx_sign` below gets the decorator directly on it instead of this subfunction?

lgeiger: Unfortunately `tf.custom_gradient` only supports key word arguments in eager mode. So we need to wrap it into the `_call` function to make it work. Checkout fa5ef93566fefecef941d33739876243527a4b79

PR #277, opened 2019-10-03
Run unit tests with GitHub actions

lgeiger:


lgeiger:
I really like the new GitHub actions experience. I think we can try it out on this repo and if we run into problems, we can roll back.

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.6
+      - name: Install dependencies
+        run: |
+          pip install tensorflow==2.0.0
```
leonoverweel: The old unit testing matrix (from `azure-pipelines.yml`) has TensorFlow 1.13.1 testing on Python 3.6, while here TensorFlow 2.0.0 runs coverage on Python 3.6. Is that right? Why isn't coverage running on Python 3.7 like before? (If I'm reading it right on line 19 of `azure-pipelines.yml`)

lgeiger: Ideally we would test all TF versions against both 3.6 and 3.7, but I think that's a bit too much. I run the coverage on Python 3.6 to have at least one safety net if we introduce Python 3.7 syntax by accident.

PR #292, opened 2019-10-16
Cleanup lq.math tests

jamescook106:
This PR touchs the math, activations and constraints test to bring them more inline with the quantizers tests.


jamescook106:
@lgeiger any advice with the failing test - it works locally for me but I am using tensorflow 2

jamescook106:
@lgeiger tests now pass

PR #302, opened 2019-10-30
Fix zero division bug in summary for weightless models

jneeven:


jneeven:
> Could you explain the bug in more detail? As I can't see it when looking at the code you have changed

If you have a model with no weights (e.g. a noop model for testing purposes), the ratio of binary weights to total weights will be a division by zero. Same for the memory size etc.

jamescook106:
@jneeven is that because ```self.fp_equivalent_memory``` goes to 0?

jneeven:
@jamescook106 exactly, `self.op_count` and `self.fp_equivalent_memory` will be 0

jamescook106:
@jneeven could be worth adding another test?

Issue #183, opened 2019-08-13
Improve customizability of Bop


lgeiger:
### Feature motivation

Currently when using the Bop optimizer, it will update the weights on all quantized layers:
https://github.com/larq/larq/blob/b32830b4cca4a69ac1daf1d176239971112620e4/larq/optimizers_v2.py#L83-L84

This can be problematic if one wants to use a quantized layer with no `kernel_quantizer` or a quantizer with higher precision.

### Feature description

It would be good to have a more fine grained control over which layers are trained with Bop and which use other precision.

### Feature implementation

One possible implementation would be to add a fake `lq.quantizers.bop` function that doesn't change the forward pass, but marks this kernel so Bop should handle weight updates. This could be achieved by using a specific name scope or adding an attribute that Bop recognises.
An other possibility would be to explicitly pass a list of layers or variables to Bop.
Can you think of a better way to handle this?

koenhelwegen:
I'm not sure if a `quantizer` is the right way to think about this. For me, a `quantizer` implies a map from a real-valued weight to a quantized weight, and so is tightly coupled to the latent-weight approach.

Ideally, the abstraction would make it obvious which approach is used (real-valued weight/latent-weight binary weight/true binary weight), facilitate conversion to an inference graph later on, and be both flexible & transparent in quantization details (bit precision, scaling factors, etc).

One (verbose and probably bad) way to achieve this is adding additional arguments `kernel_quantization_strategy`, `bias_quantization_strategy` etc. to the layers which take values like `None`, `latent`, `binary`; Bop could then target only `true_binary` variables, and we could assert the strategy is `latent` when a `quantizer` is set. We could use scopes to avoid having to set this for each layer individually; however, for models where the higher-precision layers are quantized to 8-bit a mixture of latent weights and true binary weights is actually very common.

koenhelwegen:
Another problem I have with the current API is that Bop has the real-valued optimizer as an attribute. From the users perspective it doesn't make sense to have this hierarchy - Bop & Adam exist next to each other. The current approach also quickly breaks down if you would have more than two optimizers (e.g. Bop for binary, something else for ternary and Adam for latent-weight 8-bit layers).

Not sure if there is a better approach though. Subclassing `keras.model` so that `model.compile` accepts a list of optimizers and determines which variables are targeted by which optimizer would be great but will probably make our life very complicated. A `GroupedOptimizer` class may be an idea if using three different optimizers becomes a real thing, but currently that seems pretty far out.

[...]

PR #304, opened 2019-11-01
Refactor Bop into Bop + CaseOptimizer

leonoverweel:
The aim of this refactor is to (1) take the complexity of delegating full-precision training to a full-precision optimizer out of binary optimizers like Bop; (2) make it clear to the user which predicate is being used to decide which optimizer works on which variables; and (3) set the stage for addressing the original issue in #183.

Usage example for the new optimizer API, for binarynet:

```python
@registry.register_hparams(binarynet)
class bop(default):
    @property
    def optimizer(self):
        return lq.optimizers.CaseOptimizer(
            (
                lq.optimizers.Bop.is_binary_variable,  # predicate, can be any function
                lq.optimizers.Bop(self.threshold, self.gamma),  # optimizer, must be tf.keras.optimizers.Optimizer
            ),  # (you can pass more tuples to add more optimizers)
            default=tf.keras.optimizers.Adam(0.01),
        )
```

As you can see, we introduce the concept of a `CaseOptimizer` for handling having both binary- and full-precision optimizers that are responsible for training different weights in the model. Best explained by the docs; screenshots:

image
image

**(Updated 2019-11-08)**


larq/optimizers_v2.py
```diff
@@ -5,6 +5,84 @@
 from copy import deepcopy
 
 
+@utils.register_keras_custom_object
+class OptimizerGroup(tf.keras.optimizers.Optimizer):
+    """Group of a full-precision and binary optimizer.
+
+    #TODO: Do we want to abstract this to `n` optimizers, not specifically one fp and
+    one bin? (I personally prefer the user-side simplicity of this current approach.)
```
koenhelwegen: I'm strongly in favour of the more general approach; I don't see a huge benefit in making this so specific while it removes a lot of flexibility.

larq/optimizers_v2.py
```diff
@@ -5,6 +5,84 @@
...
+            raise TypeError(type_err_msg.format("fp_optimizer", type(bin_optimizer)))
+
+        self.bin_optimizer = bin_optimizer
+        self.fp_optimizer = fp_optimizer
+
+    def apply_gradients(self, grads_and_vars, name=None):
+        bin_grads_and_vars, fp_grads_and_vars = [], []
+
+        for grad, var in grads_and_vars:
+            if self.is_binary(var):
```
koenhelwegen: I would like the decision rule to be modular, possibly even an argument given to `__init__()`, instead of hardcoded here.

larq/optimizers_v2.py
```diff
@@ -5,6 +5,84 @@
...
+
+        return tf.group(bin_train_op, fp_train_op, name="train_with_group")
+
+    def __getattr__(self, name):
+        if name == "lr":
+            return self.fp_optimizer.lr
+        return super().__getattr__(name)  # TODO is this still robust enough?
+
+    # TODO: Not sure if this needs to be here or forwarded to Bop.
+    def _get_decayed_hyper(self, name, var_dtype):
```
koenhelwegen: Yes this should be a method of the individual optimizer. I don't think any external code tries to access this method but that may be something to check.

larq/optimizers_v2.py
```diff
@@ -40,44 +119,21 @@ class Bop(tf.keras.optimizers.Optimizer):
...
         self._set_hyper("threshold", threshold)
         self._set_hyper("gamma", gamma)
 
+    # TODO: Why do we need to do this here? (It gets called in apply_gradients().)
+    # TODO: Can we remove the `is_binary` check? Will this never get called with non-
+    # binary variables anyway?
     def _create_slots(self, var_list):
         for var in var_list:
-            if self.is_binary(var):
+            if OptimizerGroup.is_binary(var):
```
koenhelwegen: `apply_gradients` only gets passed the binary vars, so there should be no additional check needed here

larq/optimizers_v2.py
```diff
@@ -5,6 +5,84 @@
 from copy import deepcopy
 
 
+@utils.register_keras_custom_object
+class OptimizerGroup(tf.keras.optimizers.Optimizer):
```
koenhelwegen: An alternative to making the implementation general is to make it clear this is a specific case, e.g. by naming the class "BNNOptimizerDuo". In this approach I would approach largely as is, on the condition that the decision rule for whether a variable is binary can be made more robust.

If the decision rule isn't robust, I think it's better if the user gets a responsibility to set the rule. Currently, passing a name to a quantized layer introduces a rather subtle bug.

(I still have a slight preference for an abstract `OptimizerGroup`, possibly with a specific `BNNOptimizerDuo` as subclass, which would trivially extend to other/more bit widths and not add much complexity in my view. But this approach would also be fine I think.)

leonoverweel: I've renamed `OptimizerGroup` to `BNNOptimizerDuo` in f65d7a1. I think we should wait to add a more general superclass around this until we have more subclasses (like a Trio with a ternary optimizer).

I agree about the subtle bugs that hiding the `_is_binary` rule in this class could introduce for users not following our binary layer name conventions; we can check back when we know whether the `tf.Variable` subclass decision rule is viable.

larq/optimizers_v2.py
leonoverweel: (See discussion [here](https://github.com/larq/larq/pull/304/files/4ce2e10b4c5ecbffc5f9ec1353f7cf5ee0a625d7#r341977897).)

larq/optimizers_v2.py
leonoverweel: Checked that this does not get called externally by TensorFlow or our code; removed from `BNNOptimizerDuo` in 3dd59a8.

larq/optimizers_v2.py
leonoverweel: Check removed in 51a20ab.

larq/optimizers_v2.py
leonoverweel: (See discussion [here](https://github.com/larq/larq/pull/304/files/4ce2e10b4c5ecbffc5f9ec1353f7cf5ee0a625d7#r341977897).)

leonoverweel:
In the last commit, I went back to the `GroupedOptimizer` approach which takes a list of `MaskedOptimizer`, each of which has a `tf.keras.optimizer.Optimizer` and a `mask_fn(var)` that decides whether that optimizer should train that variable. The optimizers are ordered, and the last one may mark itself as the default by setting its `mask_fn` to `None`.

Here's how you'd use the new API for e.g. binarynet:

```python
@registry.register_hparams(binarynet)
class bop(default):
    threshold = 1e-6
    gamma = 1e-3

    @property
    def optimizer(self):
        return lq.optimizers.OptimizerGroup(
            [
                lq.optimizers.MaskedOptimizer(
                    optimizer=lq.optimizers.Bop(self.threshold, self.gamma),
                    mask_fn=lq.optimizers.MaskedOptimizer.is_binary,
                ),
                lq.optimizers.MaskedOptimizer(
                    optimizer=tf.keras.optimizers.Adam(self.learning_rate),
                    mask_fn=None
                ),
            ]
        )
```

For more details, see the docstrings for `MaskedOptimizer` and `OptimizerGroup` in `optimizers_v2.py`.

larq/optimizers_v2.py
```diff
@@ -5,50 +5,90 @@
...
+        if not isinstance(optimizer, tf.keras.optimizers.Optimizer):
+            raise TypeError(
+                f"Expected `tf.keras.optimizers.Optimizer` for `optimizer` but got `{type(optimizer)}`."
+            )
+
+        self.optimizer = optimizer
+        self.mask_fn = mask_fn
+
+    @staticmethod
+    def is_binary(var):
```
koenhelwegen: would define this outside the class

larq/optimizers_v2.py
```diff
@@ -5,50 +5,90 @@
...
+                )
 
-        self.bin_optimizer = bin_optimizer
-        self.fp_optimizer = fp_optimizer
+        self.masked_optimizers = masked_optimizers
 
     def __getattr__(self, name):
-        if name == "lr":
-            return self.fp_optimizer.lr
+        if name == "lr":  # TODO: Which optimizer's LR should handle this?
```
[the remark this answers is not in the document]
leonoverweel: (Just throw a warning.)

larq/optimizers_v2.py
```diff
@@ -5,50 +5,90 @@
...
-        bin_grads_and_vars, fp_grads_and_vars = [], []
+        opt_grads_and_vars = [[] for _ in range(len(self.masked_optimizers))]
 
         for grad, var in grads_and_vars:
-            if self.is_binary(var):
-                bin_grads_and_vars.append((grad, var))
-            else:
-                fp_grads_and_vars.append((grad, var))
+            for i, masked_opt in enumerate(self.masked_optimizers):
+                # TODO: Check partition condition?
```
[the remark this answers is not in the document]
leonoverweel: Now being checked as of 62c28b0.

larq/optimizers_v2.py
```diff
@@ -5,88 +5,75 @@
...
+            and default is not None
+        ):
+            raise TypeError(
+                f"Expected `tf.keras.optimizers.Optimizer` for `default` but got `{type(fallback_optimizer)}`."
+            )
+        self.default = default
 
     def __getattr__(self, name):
-        if name == "lr":  # TODO: Which optimizer's LR should handle this?
+        if name == "lr":  # TODO: Return list of learning rates
```
koenhelwegen: Would raise an error, we can make this nice later if there is demand for it

larq/optimizers_v2.py
```diff
@@ -5,88 +5,75 @@
...
-        ):
-            train_ops.append(
-                masked_opt.optimizer.apply_gradients(grads_and_vars, name=name)
+            if num_opts == 0:
+                default_grads_and_vars.append((grad, var))
+            if num_opts > 1:
+                raise ValueError(f"Variable `{var}` claimed by multiple optimizers.")
+
+        if len(default_grads_and_vars) > 0 and self.default is None:
+            raise ValueError(
```
koenhelwegen: Would raise a warning, as this may be intended behaviour

larq/optimizers_v2.py
```diff
@@ -5,6 +5,124 @@
...
+        ```
+    
+    # Arguments
+    optimizer: a `tf.keras.optimizers.Optimizer`.
+    mask_fn: a function which takes only a `tf.Variable` as input and returns True if 
+        this optimizer should be used to train that variable.
+    """
+
+    def __init__(self, optimizer, mask_fn=None):
+        if not isinstance(optimizer, tf.keras.optimizers.Optimizer):
```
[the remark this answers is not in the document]
leonoverweel: Added a `callable()` check for `pred` in f1eb04f.

larq/optimizers_v2.py
leonoverweel: Implemented throwing a `ValueError` for this in 151af08.

larq/optimizers_v2.py
leonoverweel: Changed to warning in 678590c.

larq/optimizers.py
```diff
@@ -1,9 +1,9 @@
 from larq import utils
 
 if utils.tf_1_14_or_newer():
-    from larq.optimizers_v2 import Bop
+    from larq.optimizers_v2 import Bop, CaseOptimizer
 
-    __all__ = ["Bop"]
+    __all__ = ["Bop", "OptimizerGroup"]
```
lgeiger: This should be, or we need to rename `CaseOptimizer`
```suggestion
    __all__ = ["Bop", "CaseOptimizer"]
```

larq/optimizers_v2.py
```diff
@@ -1,10 +1,121 @@
...
+
+    # Arguments
+    pred_opt_pairs: a list of `(tf.keras.optimizers.Optimzer, pred)` pairs, where `pred` 
+        takes one `tf.Variable` as argument and returns `True` if the optimizer should
+        train that variable, e.g. `pred(var) === True`.
+    default: a `tf.keras.optimizers.Optimizer` to be applied to any variable not claimed
+        by any other optimizer.
+    """
+
+    def __init__(self, pred_opt_pairs, default, name="optim_case"):
```
lgeiger:
```suggestion
    def __init__(self, pred_opt_pairs, default=None, name="optimizer_case"):
```

larq/optimizers_v2.py
```diff
@@ -1,10 +1,121 @@
...
+        ):
+            raise TypeError(
+                f"Expected `tf.keras.optimizers.Optimizer` for `default` but got `{type(fallback_optimizer)}`."
+            )
+        self.default = default
+
+    def __getattr__(self, name):
+        if name == "lr":
+            raise ValueError("No single learning rate available for CaseOptimizer.")
+        return super().__getattr__(name)
```
lgeiger: Actually I don't think this will do anything since the optimizer won't have any interesting to attributes. I think it is OK to remove the whole `__getattr__` method. Or am I missing something?

larq/optimizers_v2.py
```diff
@@ -1,10 +1,121 @@
...
+
+        for grad, var in grads_and_vars:
+            num_opts = 0
+            for i, (pred, opt) in enumerate(self.pred_opt_pairs):
+                if pred(var):
+                    opt_grads_and_vars[i].append((grad, var))
+                    num_opts += 1
+
+            if num_opts == 0:
+                default_grads_and_vars.append((grad, var))
```
lgeiger: I think we should throw here if no default optimizer is defined.

larq/optimizers_v2.py
```diff
@@ -1,10 +1,121 @@
...
+            if num_opts > 1:
+                raise ValueError(f"Variable `{var}` claimed by multiple optimizers.")
+
+        train_ops = []
+        for (_, opt), grads_and_vars in zip(self.pred_opt_pairs, opt_grads_and_vars):
+            train_ops.append(opt.apply_gradients(grads_and_vars, name=name))
```
lgeiger: I always like list comprehensions at places like this, though this might not clean up the code in this case (feel free ignore this remark)

larq/optimizers_v2.py
```diff
@@ -1,10 +1,121 @@
...
+
+        if len(default_grads_and_vars) > 0 and self.default is None:
```
lgeiger: If we throw at line 68, we could remove this logic.

larq/optimizers.py
leonoverweel: Yeah oops haha, already fixed in 1b49c96 but hadn't pushed yet.

larq/optimizers_v2.py
leonoverweel: I mostly added it because Bop had it and I assumed it was getting called externally somewhere. If that's not the case I'm fine with removing it. Thoughts @koenhelwegen?

larq/optimizers_v2.py
leonoverweel: We have this a few lines later:

```python
        if len(default_grads_and_vars) > 0 and self.default is None:
            warnings.warn(
                f"No `default` provided to train variables `{default_grads_and_vars}`."
            )
```

@koenhelwegen mentioned that not having a default could be desired behavior, e.g. for the usecase of only retraining batch norms and not touching other weights. That's why we changed the error throw I had here originally to a warning.

larq/optimizers_v2.py
leonoverweel: This is how it'd look as a list comprehension:

```python
train_ops = [
    opt.apply_gradients(grads_and_vars, name=name)
    for (_, opt), grads_and_vars in zip(self.pred_opt_pairs, opt_grads_and_vars)
]
```

I'm not 100% sure which I like better myself, wdyt?

larq/optimizers_v2.py
leonoverweel: See [my comment above](https://github.com/larq/larq/pull/304#discussion_r343012326), we don't necessarily want to throw.

larq/optimizers_v2.py
leonoverweel: Another option to make it more readable, suggested by @daniel-vos:

```python
opts = [opt for (_, opt) in self.pred_opt_pairs]
train_ops = [
    opt.apply_gradients(grads_and_vars, name=name)
    for opt, grads_and_vars in zip(opts, opt_grads_and_vars)
]
```

I personally like this best; don't think the extra iteration over the optimizers matters much in terms of performance since there's not usually more than one or two optimizers.

leonoverweel: Changed in b33bd04.

larq/optimizers_v2.py
```diff
@@ -118,32 +94,34 @@ def _get_optimizer(self, var):
...
+            if num_optimizers > 1:
                 raise ValueError(f"Variable `{var}` claimed by multiple optimizers.")
-            if num_opts == 0:
+            if num_optimizers == 0:
                 self.var_opt_mapping[var.name] = self.DEFAULT_OPT_INDEX
                 if self.default is None:
                     warnings.warn(f"No `default` provided to train variable `{var}`.")
 
+            # Yielding a new grads_and_vars iterator for use by `apply_gradients()`
             yield ((grad, var))
```
koenhelwegen: This makes it somewhat confusing to grasp the execution path. We can also cast the iterator to a list (see https://github.com/tensorflow/tensorflow/issues/33345 for context).

leonoverweel: Changed in 7aa8624.

leonoverweel:
The API is now:

```python
@registry.register_hparams(binarynet)
class bop(default):
    @property
    def optimizer(self):
        return lq.optimizers.CaseOptimizer(
            [
                (
                    lq.optimizers.Bop.is_binary_variable,
                    lq.optimizers.Bop(threshold=1e-6, gamma=1e-3),
                )
            ],
            tf.keras.optimizers.Adam(0.01),
        )
```

larq/optimizers_v2.py
```diff
@@ -1,10 +1,151 @@
...
+    def apply_gradients(self, grads_and_vars, name=None):
+        """Apply gradients to variables for each optimizer.
+        
+        On the first call to `apply_gradients()`, compute the mapping from variables to
+        optimizers and cache it in the `self.var_opt_mapping` dict for serialization and
+        faster access.
+        """
+
+        if self.var_opt_mapping is None:
+            grads_and_vars = list(grads_and_vars)
```
lgeiger:
```suggestion
            # Convert `grads_and_vars` to list so we can iterate multiple times over it
            grads_and_vars = list(grads_and_vars)
```

larq/optimizers_v2.py
```diff
@@ -1,10 +1,151 @@
...
+            self._compute_var_opt_mapping(grads_and_vars)
+
+        train_ops = [
+            self._get_optimizer(var).apply_gradients([(grad, var)])
```
lgeiger: I would prefer calling `apply_gradients` for each optimizer only once, due to performance reasons.

larq/optimizers_v2.py
```diff
@@ -1,10 +1,151 @@
...
+                )
+                for opt_config in config["optimizer_configs"]
+            ],
+            default=tf.keras.optimizers.deserialize(
+                config["default_config"], custom_objects=custom_objects
+            ),
+        )
+
+        # Since we no longer have the `pred`s, we set the mapping explicitly
+        case_optimizer.var_opt_mapping = config["var_opt_mapping"]
```
lgeiger: I think this is OK to do, another way would be to pass it as `**kwargs` to the class and catch it in the `__init__`. I don't have a preference, I think both ways are equally valid.

larq/optimizers_v2.py
```diff
@@ -1,10 +1,151 @@
...
 import tensorflow as tf
 import larq as lq
 
 from larq import utils
 from copy import deepcopy
 
 
+@utils.register_keras_custom_object
+class CaseOptimizer(tf.keras.optimizers.Optimizer):
+    """A case operation for optimizers that each train a subset of a model's variables.
```
lgeiger: What do you think about:
```suggestion
    """An optmizer wrapper that applies different optimizers to a subset of variables.
```

larq/optimizers_v2.py
```diff
@@ -1,10 +1,151 @@
...
+
+    An optimizer is used to train a variable iff its accompanying predicate evaluates to
```
lgeiger: I am not a particular fan of the wording "an optimizer that trains a variable", maybe we should be a bit more precise.
```suggestion
    An optimizer is used to train a variable if its accompanying predicate evaluates to
```

larq/optimizers_v2.py
```diff
@@ -1,10 +1,151 @@
...
+    `True`.
+
+    For each variable, at most one optimizer's predicate may evaluate to `True`. If no
+    optimizer's predicate evaluates to `True` for a variable, it is trained with the
+    `default` optimizer.
+
+    # Arguments
+    pred_opt_pairs: a list of `(tf.keras.optimizers.Optimzer, pred)` pairs, where `pred`
```
lgeiger: Isn't it?
```suggestion
    pred_opt_pairs: A list of `(pred, tf.keras.optimizers.Optimzer)` pairs, where `pred` 
```

larq/optimizers_v2.py
```diff
@@ -1,10 +1,151 @@
...
+    `True`.
+
+    For each variable, at most one optimizer's predicate may evaluate to `True`. If no
+    optimizer's predicate evaluates to `True` for a variable, it is trained with the
+    `default` optimizer.
+
+    # Arguments
+    pred_opt_pairs: a list of `(tf.keras.optimizers.Optimzer, pred)` pairs, where `pred` 
+        takes one `tf.Variable` as argument and returns `True` if the optimizer should
+        train that variable, e.g. `pred(var) == True`.
```
lgeiger:
```suggestion
        be used for that variable, e.g. `pred(var) == True`.
```

larq/optimizers_v2.py
```diff
@@ -1,10 +1,151 @@
...
+    default: a `tf.keras.optimizers.Optimizer` to be applied to any variable not claimed
```
lgeiger:
```suggestion
    default: A `tf.keras.optimizers.Optimizer` to be applied to any variable not claimed
```

larq/optimizers_v2.py
```diff
@@ -97,27 +211,22 @@ def _resource_apply_dense(self, grad, var):
...
-    @staticmethod
-    def is_binary(var):
-        return "/kernel" in var.name and "quant_" in var.name
+    def _resource_apply_sparse(self, grad, var, indices):
+        raise NotImplementedError()
 
     def get_config(self):
-        fp_optimizer_config = self.fp_optimizer.get_config()
         config = {
+            "name": self._name,
```
lgeiger: This should be included in [`super().get_config()`](https://github.com/tensorflow/tensorflow/blob/6eac303dd8534e65905e379d7f42f7b4a06bebfe/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L703) already.
```suggestion
```

larq/optimizers.py
```diff
@@ -1,9 +1,41 @@
+"""Neural networks with extremely low-precision weights and activations, such as 
+Binarized Neural Networks (BNNs), usually contain a mix of low-precision and 
+higher-precision weights (e.g. 8-bit, 16-bit, or 32-bit). Examples of this include the 
+first and last layers of image classificiation models, which have higher-precision 
+weights in most BNN architectures from the literature.
```
lgeiger: Should we mention bias or batch norm variables here as well?

larq/optimizers.py
```diff
@@ -1,9 +1,41 @@
...
+
+Optimizing a BNN, then, consists of optimizing both low-precision and higher-precision
```
lgeiger:
```suggestion
Training a BNN, then, consists of optimizing both low-precision and higher-precision
```

larq/optimizers.py
```diff
@@ -1,9 +1,41 @@
...
+weights. In `larq`, we provide a mechanism to target different bit-precision variables
+with different optimizers using the `CaseOptimizer` class. Modeled after the
+[`tf.case`](https://www.tensorflow.org/api_docs/python/tf/case) signature,
+`CaseOptimizer` accepts pairs of optimizers and predicates. A predicate, given a
```
lgeiger:
```suggestion
`CaseOptimizer` accepts pairs of predicates and optimizers. A predicate, given a
```

larq/optimizers_v2.py
```diff
@@ -1,10 +1,151 @@
...
+
+    # Arguments
+    pred_opt_pairs: a list of `(tf.keras.optimizers.Optimzer, pred)` pairs, where `pred` 
+        takes one `tf.Variable` as argument and returns `True` if the optimizer should
+        train that variable, e.g. `pred(var) == True`.
+    default: a `tf.keras.optimizers.Optimizer` to be applied to any variable not claimed
+        by any other optimizer.
+    """
+
+    def __init__(self, pred_opt_pairs, default=None, name="optimizer_case"):
```
lgeiger: Should we change the signature to
```python
    def __init__(self, *pred_opt_pairs, default=None, name="optimizer_case"):
```
Then we can get rid of an extra list when using the optimizer.

larq/optimizers_v2.py
leonoverweel: This "iff" was on purpose, as in [if and only if](https://en.wikipedia.org/wiki/If_and_only_if), and not a typo. Do you prefer that or just keeping it as "if"?

What don't you like about the "an optimizer that trains a variable" phrasing exactly?

larq/optimizers_v2.py
leonoverweel: I had that initially but I slightly prefer not having it in the `__init__` since that makes it very clear that this isn't something the user should be doing by hand.

larq/optimizers_v2.py
leonoverweel: Yep, good catch. I had this switched initially before we started following the `tf.case` signature.

larq/optimizers_v2.py
leonoverweel: Nice, looks like `TestBopOptimizer.test_bop_serialization` still passes when I remove this so I'll delete it.

larq/optimizers.py
leonoverweel: Discussed this earlier in the AMS office and I don't think so, this is just a simple example of why we need variables (and optimizers) of different precisions at all, and there's no need to explain every use case for that.

larq/optimizers_v2.py
leonoverweel: I don't think this works with tuples nested inside lists; the different variations of this I tried all ended up failing for either `[(pred, opt)]` or `(pred, opt)`. If you have a way that does work please let me know.

larq/optimizers_v2.py
leonoverweel: So iterate over the `grads_and_vars` to sort them into separate lists for each optimizer first, and then passing these lists to each optimizer?

larq/optimizers_v2.py
lgeiger: :+1: Make sense, we can keep it as is.

larq/optimizers_v2.py
lgeiger: Or we can switch the mapping, from `{var: optimizer}` to `{optmizer: var}`. Not sure what's better.

I'd prefer only call `apply_gradient` once per optimizer, since we would otherwise run [this code](https://github.com/tensorflow/tensorflow/blob/6977b8f7422b436c02aafde371716319149af542/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L407-L500) many times. Not sure if this is actually a performance problem, I'd need to look deeper into that.

larq/optimizers_v2.py
lgeiger: I don't see why this would fail. The following works well for me:
```python
def foo(*pred_opt_pairs):
    print(pred_opt_pairs) 

foo((True, "opt1"), (True, "opt2"))
# --> ((True, 'opt1'), (True, 'opt2'))

foo((True, "opt1"))
# --> ((True, 'opt1'),)
```

leonoverweel: Ah, I was trying tuples nested inside lists instead of just several tuples after each other, but latter works. The only drawback here is that you always have to pass `default=` for the default, or it'll try to unpack that as a tuple too; but I think that's fine. Changed in 2ec9175.

larq/optimizers_v2.py
leonoverweel: I tried `{optimizer: var}` but it ended up being a bit messy. Made some tweaks to now precompile lists of optimizers and `grads_and_vars` for each optimizer, so that `apply_gradients` only calls each optimizer once. See 9cd01e3.

leonoverweel:
@lgeiger I think I've now implemented all your suggestions. Three small problems remain:

* TensorFlow 1.13 tests fail: should I backport this refactor to `optimizers_v1` as well? Or are we going to drop that completely?
* Python Black fails in CI for `optimizers_test.py` but I can't reproduce that locally.
* Code coverage; should I add some basic testing that touches the new class?

larq/optimizers_v2.py
```diff
@@ -1,10 +1,152 @@
...
+@utils.register_keras_custom_object
+class CaseOptimizer(tf.keras.optimizers.Optimizer):
+    """An optmizer wrapper that applies different optimizers to a subset of variables.
+
+    An optimizer is used to train a variable iff its accompanying predicate evaluates to
+    `True`.
+
+    For each variable, at most one optimizer's predicate may evaluate to `True`. If no
+    optimizer's predicate evaluates to `True` for a variable, it is trained with the
+    `default` optimizer.
```
koenhelwegen: Maybe mention that if no `default` is set, the variable is not trained?

larq/optimizers_v2.py
```diff
@@ -97,27 +216,21 @@ def _resource_apply_dense(self, grad, var):
...
-    @classmethod
-    def from_config(cls, config, custom_objects=None):
-        new_config = deepcopy(config)
-        fp_optimizer = tf.keras.optimizers.deserialize(
-            new_config["fp_optimizer"], custom_objects=custom_objects
-        )
-        new_config.pop("fp_optimizer", None)
-        return cls(fp_optimizer, **new_config)
+    @staticmethod
+    def is_binary_variable(var):
```
koenhelwegen: Maybe add example of usage or a "See also: `CaseOptimizer`"?

leonoverweel: Added in 35b3d36.

larq/optimizers_v2.py
leonoverweel: Added in 35b3d36.

larq/optimizers_test.py
```diff
@@ -111,14 +111,24 @@ def test_serialization(self):
...
 class TestBopOptimizer:
+    @pytest.mark.skipif(
+        utils.tf_1_14_or_newer() is False,
+        reason="only supporting CaseOptimizer in TensorFlow >= 1.14",
+    )
     def test_bop_accuracy(self):
         _test_optimizer(
-            lq.optimizers.Bop(fp_optimizer=tf.keras.optimizers.Adam(0.01)),
+            lq.optimizers.CaseOptimizer(
+                (lq.optimizers.Bop.is_binary_variable, lq.optimizers.Bop()),
```
lgeiger: Should we add a test for `CaseOptimizer` with multiple predicates?

larq/optimizers_v2.py
lgeiger: > This "iff" was on purpose, as in if and only if, and not a typo. Do you prefer that or just keeping it as "if"?

Got it

> What don't you like about the "an optimizer that trains a variable" phrasing exactly?

I personally view an optimizer as something that applies an update to a variable given a gradient signal, rather than training that variable, since for training it requires more things like data, a loss function, etc. But that's just a personal thing, since I never heard the term "training a single variable".

larq/optimizers_v2.py
```diff
@@ -1,10 +1,153 @@
...
+
+    # Arguments
+    pred_opt_pairs: One or more `(pred, tf.keras.optimizers.Optimzer)` pairs, where 
+        `pred`  takes one `tf.Variable` as argument and returns `True` if the optimizer 
+        should be used for that variable, e.g. `pred(var) == True`.
+    default: A `tf.keras.optimizers.Optimizer` to be applied to any variable not claimed
+        by any other optimizer. (Must be passed as keyword argument.)
+    """
+
+    def __init__(self, *pred_opt_pairs, default=None, name="optimizer_case"):
```
lgeiger: Do you think renaming `default` to `default_optimizer` would be more descriptive?

leonoverweel: `default` matches the `tf.case`, but I'd be good with calling it `default_opt` (to also match `pred_opt_pairs`). What do you think?

lgeiger: Thinking of this again, what do you think about `predicate_optimizer_pairs` and `default_optimizer`? I think it is a bit long, but would be more descriptive.

larq/optimizers_test.py
leonoverweel: I added `test_overlapping_predicates` in eeb5272.

larq/optimizers_v2.py
leonoverweel: Done in e4695d2.

leonoverweel:
@lgeiger I *think* this should be the one!

Issue #306, opened 2019-11-06
RFC: Quantized Variables


lgeiger:
This Request for Comments (RFC) outlines a proposed addition to the API of `larq`. It would be good to get feedback on this, so please ask questions if I need to elaborate more in some areas. WIP implementation can be viewed at #312.

## Objective

`larq` has a concept of quantized layers that define a [quantized operation using quantized inputs and variables](https://larq.dev/guides/key-concepts/#quantized-layers). Quantized layers are currently the only official way to interact with `larq`. E.g. accessing the precision of quantized variables or inputs are cumbersome to get and require checking `layer.quantized_latent_weights` and `layer.quantizers`. This often requires downstream code that needs access to variables to rely on private methods. We can fix this by either introducing convenience functions and a stable API of quantized layers to access quantized variables or extend the API with the concept of a quantized variable. This document will elaborate on the latter since it allows new use cases for `larq` and gives users more fine grained control when using custom layers or no layers at all. This concept can be extended to `tf.Tensor` for quantized activations which we can discuss in a future RFC.

## Design Proposal

There are a few ways we could implement this functionality, for the initial proposal I closely followed the implementation of [`AutoCastVariable`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/mixed_precision/experimental/autocast_variable.py#L28) which only wraps `tf.Variable`.
While the variable acts like a quantized variable the TensorFlow data type will still be floating-point in order to be compatible with all TensorFlow operations. This is typically referred to as fake quantization.

```python
class QuantizedVariable(tf.Variable):
    def __init__(self, variable, quantizer=None, precision=None):
        """Creates an QuantizedVariable instance.

        # Arguments
        variable: A floating-point resource variable to wrap.
        quantizer: An optional quantizer to transform the floating-point
            variable to a fake quantized variable.
        precision: An optional integer defining the precision of the quantized
            variable. If `None`, `quantizer.precision` is used.
        """
        pass

    @property
    def precision(self):
        """Returns the integer precision of the quantized variable."""
        pass

    def latent_value(self):
        """Returns the latent value of the variable."""
        pass

    # Overload all operators to use the quantized version and delegate all assign
    # operations to the latent weight.
```

[...]

PR #312, opened 2019-11-07
RFC: Quantized Variables

lgeiger:
This is an initial implementation of #306 to explore the idea further.

## TODO
- [x] Figure out how to deal with variable assignment: https://github.com/tensorflow/tensorflow/issues/34332, https://github.com/tensorflow/tensorflow/pull/34779, fixed in f44f24eec8785015360db2043467c6cf19a1ed83
- [x] Add unit tests for `QuantizedVariable` and manually check special cases
  - [x] Test operator overloads (e.g. `__add__`, `__mul__`)
  - [x] Test Multi GPU training with `tf.DistributedVariable`
  - [x] Test Keras Float 16 training using `AutoCastVariable`
  - [x] Test Model checkpointing and saving, should be working correctly now
  - [x] Test Gradient computation in eager, functional and legacy graph mode
  - [x] Fix unit tests for TF 1.14
  - [x] Train CIFAR10 model to verify end-to-end workflow
    - [x] TensorFlow 2.0
    - [x] TensorFlow 1.14

### DIFF
--- a/larq/conftest.py
+++ b/larq/conftest.py
@@ -3,6 +3,8 @@ import tensorflow as tf
 from packaging import version
 from tensorflow.python.eager import context
 
+from larq import quantized_scope
+
 
 @pytest.fixture
 def eager_mode():
@@ -61,4 +63,11 @@ def distribute_scope(request):
         with tf.distribute.MirroredStrategy(["cpu:0"]).scope():
             yield request.param
     else:
+        yield request.param
+
+
+@pytest.fixture(params=[True, False])
+def quantized(request):
+    """pytest fixture for running test quantized and non-quantized"""
+    with quantized_scope.scope(request.param):
         yield request.param
--- a/larq/conftest.py
+++ b/larq/conftest.py
@@ -46,3 +46,12 @@ def keras_should_run_eagerly(request):
     else:
         with context.eager_mode():
             yield request.param == "tf_keras_eager"
+
+
+@pytest.fixture(params=["default", "distribute"])
+def distribute_scope(request):
+    if request.param == "distribute":
+        with tf.distribute.MirroredStrategy(["cpu:0"]).scope():
+            yield request.param
+    else:
+        yield request.param
--- a/larq/conftest.py
+++ b/larq/conftest.py
@@ -14,14 +14,20 @@ def eager_mode():
 def graph_mode():
     """pytest fixture for running test in graph mode"""
     with context.graph_mode():
-        yield
+        with tf.compat.v1.Session().as_default():
+            yield
 
 
 @pytest.fixture(params=["eager", "graph"])
 def eager_and_graph_mode(request):
     """pytest fixture for running test in eager and graph mode"""
-    with getattr(context, f"{request.param}_mode")():
-        yield request.param
+    if request.param == "graph":
+        with context.graph_mode():
+            with tf.compat.v1.Session().as_default():
+                yield request.param
+    else:
+        with context.eager_mode():
+            yield request.param
 
 
 @pytest.fixture(params=["graph", "tf_eager", "tf_keras_eager"])
--- a/larq/conftest.py
+++ b/larq/conftest.py
@@ -1,5 +1,6 @@
 import pytest
 import tensorflow as tf
+from packaging import version
 from tensorflow.python.eager import context
 
 
@@ -44,7 +45,7 @@ def keras_should_run_eagerly(request):
     """
 
     if request.param == "graph":
-        if int(tf.__version__[0]) >= 2:
+        if version.parse(tf.__version__) >= version.parse("2"):
             pytest.skip("Skipping graph mode for TensorFlow 2+.")
 
         with context.graph_mode():
--- a/larq/conftest.py
+++ b/larq/conftest.py
@@ -48,9 +48,9 @@ def keras_should_run_eagerly(request):
             yield request.param == "tf_keras_eager"
 
 
-@pytest.fixture(params=["default", "distribute"])
+@pytest.fixture(params=[False, True])
 def distribute_scope(request):
-    if request.param == "distribute":
+    if request.param is True:
         with tf.distribute.MirroredStrategy(["cpu:0"]).scope():
             yield request.param
     else:
--- a/larq/layers_base.py
+++ b/larq/layers_base.py
@@ -66,9 +66,7 @@ class QuantizerBase(BaseLayer):
             )
 
     def get_quantizer(self, name):
-        if name == "kernel":
-            return self.kernel_quantizer
-        return None
+        return self.kernel_quantizer if name == "kernel" else None
 
     def build(self, input_shape):
         super().build(input_shape)
@@ -123,9 +121,7 @@ class QuantizerDepthwiseBase(BaseLayer):
             )
 
     def get_quantizer(self, name):
-        if name == "depthwise_kernel":
-            return self.depthwise_quantizer
-        return None
+        return self.depthwise_quantizer if name == "depthwise_kernel" else None
 
     def build(self, input_shape):
         super().build(input_shape)
--- a/larq/layers_base.py
+++ b/larq/layers_base.py
@@ -3,7 +3,7 @@ import logging
 import tensorflow as tf
 
 from larq import metrics as lq_metrics
-from larq import quantized_variable, quantizers
+from larq import quantized_scope, quantized_variable, quantizers
 
 log = logging.getLogger(__name__)
 
@@ -82,9 +82,10 @@ class QuantizerBase(BaseLayer):
     def call(self, inputs):
         if self.input_quantizer:
             inputs = self.input_quantizer(inputs)
-        if hasattr(self, "flip_ratio"):
-            self.add_metric(self.flip_ratio(self.kernel))
-        return super().call(inputs)
+        with quantized_scope.scope(True):
+            if hasattr(self, "flip_ratio"):
+                self.add_metric(self.flip_ratio(self.kernel))
+            return super().call(inputs)
 
     def get_config(self):
         config = {
@@ -138,10 +139,10 @@ class QuantizerDepthwiseBase(BaseLayer):
     def call(self, inputs):
         if self.input_quantizer:
             inputs = self.input_quantizer(inputs)
-
-        if hasattr(self, "flip_ratio"):
-            self.add_metric(self.flip_ratio(self.depthwise_kernel))
-        return super().call(inputs)
+        with quantized_scope.scope(True):
+            if hasattr(self, "flip_ratio"):
+                self.add_metric(self.flip_ratio(self.depthwise_kernel))
+            return super().call(inputs)
 
     def get_config(self):
         config = {
@@ -230,12 +231,12 @@ class QuantizerSeparableBase(BaseLayer):
     def call(self, inputs):
         if self.input_quantizer:
             inputs = self.input_quantizer(inputs)
-
-        if hasattr(self, "depthwise_flip_ratio"):
-            self.add_metric(self.depthwise_flip_ratio(self.depthwise_kernel))
-        if hasattr(self, "pointwise_flip_ratio"):
-            self.add_metric(self.pointwise_flip_ratio(self.pointwise_kernel))
-        return super().call(inputs)
+        with quantized_scope.scope(True):
+            if hasattr(self, "depthwise_flip_ratio"):
+                self.add_metric(self.depthwise_flip_ratio(self.depthwise_kernel))
+            if hasattr(self, "pointwise_flip_ratio"):
+                self.add_metric(self.pointwise_flip_ratio(self.pointwise_kernel))
+            return super().call(inputs)
 
     def get_config(self):
         config = {
--- a/larq/layers_base.py
+++ b/larq/layers_base.py
@@ -22,9 +22,9 @@ class BaseLayer(tf.keras.layers.Layer):
         if quantizer is None:
             return super()._add_variable_with_custom_getter(name, **kwargs)
 
-        # Wrap 'getter' with a version that returns an QuantizedVariable.
         old_getter = kwargs.pop("getter")
 
+        # Wrap `getter` with a version that returns a `QuantizedVariable`.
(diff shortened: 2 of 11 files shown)


larq/models.py
```diff
@@ -230,14 +230,9 @@ def generate_table_row(self, table_config):
 
     def _get_bitwidth(self, weight):
         try:
-            for quantizer, quantized_weight in zip(
-                self._layer.quantizers, self._layer.quantized_latent_weights
-            ):
-                if quantized_weight is weight:
-                    return quantizer.precision
+            return weight.precision
         except:
```
leonoverweel:
```suggestion
         except AttributeError:
```

Assuming this'd be an `AttributeError`, I think it's better to catch it more explicitly.

leonoverweel: Or maybe:

```python 
    def _get_bitwidth(self, weight):
        return getattr(weight.precision, 32)
```

larq/optimizers_test.py
```diff
@@ -1,6 +1,5 @@
 import pytest
 import numpy as np
-import tensorflow as tf
```
leonoverweel: Why this change? In most of the codebase we make calls to `tf.keras.something`, not `keras.something`, right?

larq/layers_base.py
```diff
@@ -1,14 +1,46 @@
...
-class QuantizerBase(tf.keras.layers.Layer):
+class BaseLayer(tf.keras.layers.Layer):
+    """Base class for defining quantized layers"""
+
+    def get_quantizer(self, name):
+        return None
+
+    def add_weight(self, name=None, **kwargs):
+        quantizer = self.get_quantizer(name)
+        if quantizer is not None:
```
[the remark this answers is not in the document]
lgeiger: Yes, we should revisit this function and see if we really need to rely on the private `base_layer_utils.make_variable` method. This is mainly copy paste from TensorFlow code.

larq/layers_base.py
```diff
@@ -1,14 +1,46 @@
 import logging
 import tensorflow as tf
-from larq import quantizers, utils, metrics as lq_metrics
+from larq import quantizers, quantized_variable, metrics as lq_metrics
+from tensorflow.python.keras.engine import base_layer_utils
 
 log = logging.getLogger(__name__)
 
 
 # TODO: find a good way remove duplication between QuantizerBase, QuantizerDepthwiseBase and QuantizerSeparableBase
```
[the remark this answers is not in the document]
lgeiger: We're not quite there due to the metrics, but `BaseLayer` is a start.

larq/optimizers_test.py
lgeiger: > Why this change?

No reason, mainly caused by some careless early morning debugging. 👍 Reverted the change

larq/layers_base.py
```diff
@@ -34,35 +64,25 @@ def __init__(
                 "may result in starved weights (where the gradient is always zero)."
             )
 
+    def get_quantizer(self, name):
+        if name == "kernel":
+            return self.kernel_quantizer
+        return None
```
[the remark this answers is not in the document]
lgeiger: Me too

larq/models.py
lgeiger: Fixed in 2f79a608529764e6c7bd944deeb44aa4268155c7

larq/layers_base.py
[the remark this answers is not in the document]
leonoverweel: @lgeiger I probably shouldn't have put this as suggested change because the 

```python
        if name == "kernel":
            return self.kernel_quantizer
```

is still there now.

larq/quantized_variable.py
```diff
@@ -0,0 +1,290 @@
+"""Contains QuantizedVariable, a variable that can be quantized in the forward pass."""
+from tensorflow.python.distribute import values as distribute_values
+from tensorflow.python.eager import context
+from tensorflow.python.framework import ops
+from tensorflow.python.ops import resource_variable_ops
+from tensorflow.python.ops import variables
+from functools import wraps
+
+
+def quantize(method):
```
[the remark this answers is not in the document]
lgeiger: We could also do it the same way we deal with operators and and keep a list of methods we want to delegate to the `latent_variable` and methods where we want to return the quantized version and then use `setattr` to set the method on the class.

Though I think the current way is a lot simpler, although it results in a lot more code.

larq/layers_base.py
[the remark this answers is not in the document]
lgeiger: Done in 3ce44e88ffcf4ef6f49a22dc382022f5e8e7c4cf

lgeiger:
I think this is ready to take for a spin. I added a TODO list above for the things that still need to be done before merging.

larq/quantized_variable.py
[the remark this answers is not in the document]
lgeiger: I had to remove the nice decorator in fe195dd5a26cff390dba04bbfb4fc3d3125a9555 since I forgot about some other cases we need to handle.

lgeiger:
This is ready for a first round of reviews. I'll add more unittesting on the flight to NeurIPS.

larq/conftest.py
```diff
@@ -14,14 +14,20 @@ def eager_mode():
 def graph_mode():
     """pytest fixture for running test in graph mode"""
     with context.graph_mode():
-        yield
+        with tf.compat.v1.Session().as_default():
```
leonoverweel: What is this for?

larq/conftest.py
```diff
@@ -46,3 +52,12 @@ def keras_should_run_eagerly(request):
     else:
         with context.eager_mode():
             yield request.param == "tf_keras_eager"
+
+
+@pytest.fixture(params=[False, True])
```
leonoverweel: It might be nicer to make these params strings describing what the different settings are instead of booleans, mirroring previous fixtures:
- `eager_and_graph_mode` (`["eager", "graph"]`) and
- `keras_should_run_eagerly ` (`["graph", "tf_eager", "tf_keras_eager"]`)

larq/layers_base.py
```diff
@@ -3,15 +3,47 @@
...
+        quantizer = self.get_quantizer(name)
+        if quantizer is None:
+            return super()._add_variable_with_custom_getter(name, **kwargs)
+
+        old_getter = kwargs.pop("getter")
+
+        # Wrap `getter` with a version that returns a `QuantizedVariable`.
+        def getter(*args, **kwargs):
+            variable = old_getter(*args, **kwargs)
+            return quantized_variable.create_quantized_variable(variable, quantizer)
```
leonoverweel: What do you think of doing this as a [factory pattern](https://pythonspot.com/factory-method/)? The call would then be:

```python
return quantized_variable.QuantizedVariable.factory(variable, quantizer)
```

This may just be my Java background talking though 😅

larq/quantized_variable.py
```diff
@@ -0,0 +1,308 @@
...
+        # Arguments
+        variable: A floating-point resource variable to wrap.
+        quantizer: An optional quantizer to transform the floating-point
+            variable to a fake quantized variable.
+        precision: An optional integer defining the precision of the quantized
+            variable. If `None`, `quantizer.precision` is used.
+        """
+        if not resource_variable_ops.is_resource_variable(variable):
+            raise ValueError(
+                f"variable must be of type tf.ResourceVariable, but got: {variable}"
```
leonoverweel: Do we want to report `type(variable)` here?

larq/quantized_variable.py
```diff
@@ -0,0 +1,308 @@
...
+            )
+        if not (quantizer is None or callable(quantizer)):
+            raise ValueError(
+                f"quantizer must be callable or None, but got: {quantizer}"
+            )
+        if not (precision is None or type(precision) == int):
+            raise ValueError(
+                f"precision must be of type integer or None, but got: {precision}"
```
leonoverweel: Also report `type(precision)` here?

larq/quantized_variable.py
```diff
@@ -0,0 +1,308 @@
...
+        precision: An optional integer defining the precision of the quantized
+            variable. If `None`, `quantizer.precision` is used.
+        """
+        if not resource_variable_ops.is_resource_variable(variable):
+            raise ValueError(
+                f"variable must be of type tf.ResourceVariable, but got: {variable}"
+            )
+        if not (quantizer is None or callable(quantizer)):
+            raise ValueError(
+                f"quantizer must be callable or None, but got: {quantizer}"
```
leonoverweel: Also report `type(quantizer)` here?

larq/quantized_variable.py
```diff
@@ -0,0 +1,308 @@
...
+        return self._quantize(self.latent_variable).numpy()
+
+    def sparse_read(self, *args, **kwargs):
+        return self._quantize(self.latent_variable.sparse_read(*args, **kwargs))
+
+    def gather_nd(self, *args, **kwargs):
+        return self._quantize(self.latent_variable.gather_nd(*args, **kwargs))
+
+    def __getattr__(self, name):
+        return getattr(self.latent_variable, name)
```
leonoverweel: Following [this SO post](https://stackoverflow.com/a/1944708/1814949), should we do error handling for nonexistent names here?

larq/quantized_variable.py
```diff
@@ -0,0 +1,308 @@
...
+
+
+QuantizedVariable._OverloadAllOperators()
+tf.register_tensor_conversion_function(
+    QuantizedVariable, QuantizedVariable._dense_var_to_tensor
+)
+ops.register_dense_tensor_like_type(QuantizedVariable)
+
+
+def create_quantized_variable(variable, quantizer=None, precision=None):
```
leonoverweel: See above; I think this would make sense as a factory within the `QuantizedVariable` class.

larq/quantized_variable.py
```diff
@@ -0,0 +1,308 @@
...
+    if wrap and resource_variable_ops.is_resource_variable(variable):
+        return create_quantized_variable(variable, quantizer, precision)
+    return variable
+
+
+def _get_name(obj):
+    try:
+        return obj.__name__
+    except AttributeError:
+        return obj.__class__.__name__
```
leonoverweel: Since this is just used in `QuantizedVariable`, I think it may be better to make it a static internal method of that class.

larq/quantized_variable_test.py
```diff
@@ -0,0 +1,279 @@
...
+
+def get_var(val, dtype=None, name=None):
+    return tf.compat.v1.Variable(val, use_resource=True, dtype=dtype, name=name)
+
+
+@pytest.fixture(params=[True, False])
+def quantized(request):
+    """pytest fixture for running test quantized and non-quantized"""
+    with quantized_scope.scope(request.param):
+        yield request.param
```
leonoverweel: I think using strings instead of bools would be nicer here too.

leonoverweel: And maybe we can move this into `conftest.py`, where most other pytest fixtures are. (There's only one other fixture that's not in that file: one in `models_test.py`; I think moving them all to `conftest.py` to have one place to look for fixtures would make sense.)

larq/conftest.py
lgeiger: This is so that [`evaluate`](https://github.com/larq/larq/pull/312/files#diff-915b33c29d82e4e01202fe2566e0f20fR26) can evaluate the tensor in a shared session per unittest.

larq/conftest.py
lgeiger: I opted for booleans so it is easy to check whether we are running the unittest distributed (see [here](https://github.com/larq/larq/pull/312/files#diff-2c6d294a8f4ae0b3b2ee3df2966dca03R31))

larq/layers_base.py
lgeiger: Hmm I opted for this since it is the way `AutoCastVariable` handles it is pretty simple. I'd be fine with something like `QuantizedVariable.from_variable` though. Since we currently don't expose the `QuantizedVariable` to the user by default, we can also change this in the future.

larq/quantized_variable.py
lgeiger: Yeah might be a good thing. I only copied this input validation from `AutoCastVariable` without thinking to much about it ;)

larq/quantized_variable.py
lgeiger: Not sure. I think it's fine not to handle any errors here, since `latent_variable` variable will throw an readable error message  if the attribute doesn't exist.

larq/quantized_variable.py
lgeiger: How do you feel about violating PEP8 and naming this function `QuantizedVariable` so it feels like a class when using it?

larq/quantized_variable.py
lgeiger: :+1: Makes sense.

larq/quantized_variable_test.py
lgeiger: I think I like returning booleans since it makes using the fixture in the test very easy since I only have to do `if quantized:` and not check if the return value matches a string. I put it here since this is the only file we need to use it.

larq/conftest.py
leonoverweel: OK makes sense.

larq/layers_base.py
leonoverweel: OK I'm fine with that as well.

larq/quantized_variable.py
leonoverweel: It feels a bit wrong but I think it might make sense here.

larq/quantized_variable_test.py
leonoverweel: OK booleans are good with me.

For the fixture location: we also have the `distribute_scope` and `eager_and_graph_mode` fixtures in `conftest.py` even though they only get used by `quantized_variable_test.py`, so I think we should treat the `quantized` fixture the same. We could then also test it in `conftest_test.py` like the other two.

larq/quantized_variable.py
lgeiger: > It feels a bit wrong but I think it might make sense here.

Yeah I also feel this would be kinda wrong

PR #313, opened 2019-11-08
Add pytest fixture to run tests in eager and graph mode

lgeiger:
This should allow us to get rid of `tf.test.TestCase` and allow us to fully rely on `pytest` to execute tests in eager and graph mode. `tf.test.TestCase` doesn't play well with more advanced features of `pytest` so it makes sense to consolidate our testing setup to fully rely on `pytest`.

### DIFF
--- a/larq/conftest.py
+++ b/larq/conftest.py
@@ -1,0 +1,23 @@
+import pytest
+from tensorflow.python.eager import context
+
+
+@pytest.fixture
+def eager_mode():
+    """pytest fixture for running test in eager mode"""
+    with context.eager_mode():
+        yield
+
+
+@pytest.fixture
+def graph_mode():
+    """pytest fixture for running test in graph mode"""
+    with context.graph_mode():
+        yield
+
+
+@pytest.fixture(params=["eager", "graph"])
+def eager_and_graph_mode(request):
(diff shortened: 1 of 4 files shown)


larq/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+    """pytest fixture for running test in graph mode"""
+    with context.graph_mode():
+        yield
+
+
+@pytest.fixture(params=["eager", "graph"])
+def eager_and_graph_mode(request):
+    """pytest fixture for running test in eager and graph mode"""
+    with getattr(context, f"{request.param}_mode")():
+        yield request.param
```
Tombana: Can you explain how this works?
What is the `request` parameter to `eager_and_graph_mode`, where does that parameter come from?

In the `conftest_test.py` I see a line `eager_and_graph_mode == "eager"`. Does that line implicitly make a call to this function?

lgeiger: This was taken from the [`pytest` fixtures example](https://docs.pytest.org/en/latest/fixture.html#parametrizing-fixtures).

It will run the test twice once inside `context.eager_mode()` and once inside `context.graph_mode()`. `request.param` will either be `eager` or `graph` and its value is then passed to the test.

Issue #314, opened 2019-11-08
Refactor layer test to use pytest parameterized test


lgeiger:
Currently we use a mix of unittests written in `pytest` style and tests using `tf.test.TestCase` in larq. In #313 I added some simple `pytest` fixtures that allow to run tests in both eager and graph mode.

The only part of the test suite that uses `tf.test.TestCase` are the [layer tests](https://github.com/larq/larq/blob/f6f9277a006f6cb07b1e5f9bc591a087f2261152/larq/layers_test.py#L73-L187) which rely on [`keras_parameterized.run_all_keras_modes`](https://github.com/tensorflow/tensorflow/blob/669080135924d204a1c67dd556223d794efbf664/tensorflow/python/keras/keras_parameterized.py#L176-L303)  and a forked/copied version of [`testing_utils. layer_test`](https://github.com/tensorflow/tensorflow/blob/669080135924d204a1c67dd556223d794efbf664/tensorflow/python/keras/testing_utils.py#L75-L261).

I think it would be a lot cleaner to write a simple [`pytest` fixture](https://pytest.org/en/latest/fixture.html#fixture) similar to #313 that allows us to run a normal [`pytest` parameterized test](https://pytest.org/en/latest/parametrize.html#pytest-mark-parametrize) in all Keras modes and adapt the layer tests to use it. That way we can get rid of `absl.testing` and some requirements on TensorFlow internals.

PR #317, opened 2019-11-11
Drop TensorFlow 1.13 

### LINKED ISSUE #101: Add support for XavierLarningRateScaling in TF 1.14+

leonoverweel:
This PR drops support for TensorFlow 1.13 because maintaining proper support represents too much overhead (a number of small issues + the optimizer refactor in #304). Some specifics:

* Remove `tf_1_14_or_newer()` and `supports_metrics()` definitions and uses; everything is now supported at 1.14+.
* Delete `optimizers_v1.py` and `optimizers_v2.py` and move the contents of the latter into `optimizer.py`.
* Delete `XavierScaling`, which was only implemented in `optimizers_v1.py`; closes #101 since we're no longer adding support for Xavier scaling.
* Remove a bunch of `try/except` statements related to 1.13 support.


larq/metrics.py
```diff
@@ -71,7 +66,7 @@ def get_training_metrics():
 
 
 class LarqMetric(Metric):
```
lgeiger: I think we can actually completely get rid of this class, since its only use is to add support for the `dtype ` argument in `add_weight` on TF 1.13.

larq/quantizers.py
```diff
@@ -50,12 +50,7 @@
 
 import tensorflow as tf
 from larq import utils, math
-
-try:
-    from tensorflow import is_tensor
-except:  # pragma: no cover
-    # For compatibility with TensorFlow 1.13
-    from tensorflow.python.framework.tensor_util import is_tensor
+from tensorflow import is_tensor
```
lgeiger: This import is only there due to TF 1.13 compatibility. I'd prefer to use `tf.is_tensor` instead.

leonoverweel: Yeah that's cleaner; changed in 310ace2.

larq/testing_utils.py
```diff
@@ -150,15 +149,13 @@ def layer_test(
     # See b/120160788 for more details. This should be mitigated after 2.0.
     model = tf.keras.models.Model(x, layer(x))
     if _thread_local_data.run_eagerly is not None:
-        # TensorFlow < 1.14 + eager is broken, skip the test
```
lgeiger: This entire function is just [forked from Tensorflow](https://github.com/tensorflow/tensorflow/blob/3400bf99b9cda9603db18680435091f595c4f6c5/tensorflow/python/keras/testing_utils.py#L75) which _might_ be due to TF 1.13 incompatibility. We can check if it makes sense to remove it in a future PR. Although this might conflict with #314

larq/metrics.py
leonoverweel: Nice; removed it in 6c9c691. Also checked and made sure that the signature and argument defaults still match [`tf.keras.metrics.Metric.add_weight()`](https://github.com/tensorflow/tensorflow/blob/r2.0/tensorflow/python/keras/metrics.py#L245) and they do.

PR #319, opened 2019-11-12
Refactor tests to not use TensorFlow TestCase

### LINKED ISSUE #314: Refactor layer test to use pytest parameterized test

leonoverweel:
This consolidates our tests to all use `pytest.fixture` and `pytest.mark.parameterize`. Closes #314.

### DIFF
--- a/larq/layers_test.py
+++ b/larq/layers_test.py
@@ -159,7 +159,7 @@ class LayersTest(keras_parameterized.TestCase):
 
 
 # TODO: Move back into class
-def test_depthwise_layers(fixture_run_all_keras_modes):
+def test_depthwise_layers(keras_should_run_eagerly):
     input_data = random_input((2, 3, 7, 6))
     random_weight = np.random.random() - 0.5
 
@@ -173,7 +173,7 @@ def test_depthwise_layers(fixture_run_all_keras_modes):
                 depthwise_initializer=tf.keras.initializers.constant(random_weight),
             ),
             input_data=input_data,
-            should_run_eagerly=fixture_run_all_keras_modes,
+            should_run_eagerly=keras_should_run_eagerly,
         )
 
     fp_output = testing_utils.layer_test(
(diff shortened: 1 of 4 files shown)


larq/layers_test.py
```diff
@@ -87,6 +86,7 @@ def test_binarization(self, quantized_layer, layer, input_shape, kwargs):
                     kernel_initializer=tf.keras.initializers.constant(random_weight),
                 ),
                 input_data=input_data,
+                should_run_eagerly=keras_should_run_eagerly,
             )
 
         fp_output = testing_utils.layer_test(
```
lgeiger: At some point we should get rid of this call to `layer_test` as well.

larq/testing_utils.py
```diff
@@ -2,9 +2,6 @@
 import numpy as np
 import tensorflow as tf
 
-# We should find a better solution without relying on private objects
-from tensorflow.python.keras.testing_utils import _thread_local_data, should_run_eagerly
```
lgeiger: Nice, I was worried that this will just brake at some point. Glad we can get rid of it.

PR #322, opened 2019-11-14
Quantizer layer tests

leonoverweel:
This PR adds tests for using objects that inherit from `tf.keras.layers.Layer` as input or kernel quantizer. It also moves `random_input()` from `layers_test.py` to `testing_utils.py` so that `quantizer_test.py` can also use it.

### DIFF
--- a/larq/quantizers_test.py
+++ b/larq/quantizers_test.py
@@ -61,12 +61,12 @@ class TestCommonFunctionality:
             lq.quantizers.get("unknown")
 
     @pytest.mark.parametrize("quantizer", ["input_quantizer", "kernel_quantizer"])
-    def test_layer_as_quantizer(self, quantizer):
+    def test_layer_as_quantizer(self, quantizer, keras_should_run_eagerly):
         input_data = testing_utils.random_input((1, 10))
         model = tf.keras.Sequential(
             [lq.layers.QuantDense(1, **{quantizer: DummyTrainableQuantizer()})]
         )
-        model.compile(optimizer="sgd", loss="mse")  # TODO modes?
+        model.compile(optimizer="sgd", loss="mse", run_eagerly=keras_should_run_eagerly)
         model.fit(input_data, np.ones((1,)), epochs=1)
 
 
--- a/larq/quantizers_test.py
+++ b/larq/quantizers_test.py
@@ -3,7 +3,7 @@ import numpy as np
 
 import pytest
 import larq as lq
-from larq.testing_utils import generate_real_values_with_zeros
+from larq import testing_utils
 
 
 class TestCommonFunctionality:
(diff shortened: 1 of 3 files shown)


leonoverweel:
Hmmm, TensorFlow 1.14 now fails `TestLayerWarns.test_layer_does_not_warn` because of a deprecation warning being in the logs when it checks `assert caplog.records == []`. Seems odd that this starts showing up now, right?

larq/quantizers_test.py
```diff
@@ -49,6 +60,22 @@ def test_invalid_usage(self):
...
+        model.compile(optimizer="sgd", loss="mse")  # TODO modes?
+        model.fit(input_data, np.ones((1,)), epochs=1)
+
+    def test_layer_as_kernel_quantizer(self):
+        input_data = testing_utils.random_input((1, 10))
+        model = tf.keras.Sequential(
+            [lq.layers.QuantDense(1, kernel_quantizer=DummyTrainableQuantizer())]
+        )
+        model.compile(optimizer="sgd", loss="mse")  # TODO modes?
+        model.fit(input_data, np.ones((1,)), epochs=1)
```
lgeiger: I think it would be good to check if `dummy_weight` is in `model.trainable_variables` here.

larq/quantizers_test.py
```diff
@@ -49,6 +60,22 @@ def test_invalid_usage(self):
...
+
+    def test_layer_as_kernel_quantizer(self):
+        input_data = testing_utils.random_input((1, 10))
+        model = tf.keras.Sequential(
+            [lq.layers.QuantDense(1, kernel_quantizer=DummyTrainableQuantizer())]
+        )
+        model.compile(optimizer="sgd", loss="mse")  # TODO modes?
```
lgeiger: I am fine with this. We could run compile the model eagerly, but I think that might be overkill. On the other hand I would also never say no to an additional unittest, if it is easy to include.

larq/quantizers_test.py
```diff
@@ -49,6 +60,22 @@ def test_invalid_usage(self):
...
 
+    def test_layer_as_input_quantizer(self):
+        input_data = testing_utils.random_input((1, 10))
+        model = tf.keras.Sequential(
+            [lq.layers.QuantDense(1, input_quantizer=DummyTrainableQuantizer())]
+        )
+        model.compile(optimizer="sgd", loss="mse")  # TODO modes?
+        model.fit(input_data, np.ones((1,)), epochs=1)
+
+    def test_layer_as_kernel_quantizer(self):
```
lgeiger: Should we use a parametrized test here?

leonoverweel: Combined and parameterized in a76913a.

larq/quantizers_test.py
leonoverweel: It was simple to add; added in 2758d13.

larq/quantizers_test.py
leonoverweel: Added in f44f187.

PR #328, opened 2019-11-19
Improve linting

AdamHillier:


setup.py
```diff
@@ -27,9 +27,13 @@ def readme():
         "tensorflow": ["tensorflow>=1.14.0"],
         "tensorflow_gpu": ["tensorflow-gpu>=1.14.0"],
         "test": [
+            "black==19.10b0",
+            "flake8~=3.7.9",
+            "isort~=4.3.21",
             "pytest==5.2.4",
             "pytest-cov==2.8.1",
             "pytest-xdist==1.30.0",
```
lgeiger: Maybe we should start using `~=` for those dependencies as well.

AdamHillier:
@lgeiger do you think it would make sense test on every combination of Python/TF, [like in Zookeeper](https://github.com/larq/zookeeper/blob/b1873ffc0904767d4377dd317ea16737f7708809/.github/workflows/unittest.yml#L13-L15), and then we won't have to have a separate code-coverage job, because we could do [this](https://github.com/larq/zookeeper/blob/b1873ffc0904767d4377dd317ea16737f7708809/.github/workflows/unittest.yml#L30)?

lgeiger:
> @lgeiger do you think it would make sense test on every combination of Python/TF, like in Zookeeper, and then we won't have to have a separate code-coverage job, because we could do this?

I don't really have an opinion on this. I'd be OK with testing every possible combination since test are now running parallel and don't take a lot of time. Though to me the Python 3.6 test is mostly to sanity check that we don't introduce incompatible syntax.

AdamHillier:
Okay that leads to quite a lot of GH actions so I'm going to revert it and leave it be

PR #332, opened 2019-11-20
Fix try except in model summary

lgeiger:
This was introduced in #328 and now makes model summary fail for some models since `x` in some cases can be `?` which makes `np.isnan` throw in a way that isn't caught by `AssertionError`.


AdamHillier:
Ah oops, sorry about that

PR #334, opened 2019-11-22
Fix reloading of Bop when used with schedules

lgeiger:
This will fix deserialisation of Bop when used with learning rate schedules. The fix is following the way [TensorFlow handles optimizer reloading](https://github.com/tensorflow/tensorflow/blob/8ba91e00ca8ee525279094ba8b623b2b83489996/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L711-L733).

### DIFF
--- a/larq/optimizers.py
+++ b/larq/optimizers.py
@@ -270,6 +270,15 @@ class Bop(tf.keras.optimizers.Optimizer):
         }
         return {**super().get_config(), **config}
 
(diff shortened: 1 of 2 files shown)


lgeiger:
@koenhelwegen We are not really sure how this worked before (even before the case optimizer changes) so you might want to check your model reloading code.

PR #356, opened 2019-12-16
Make HyperparameterScheduler compatible with CaseOptimizer

### LINKED ISSUE #347: Make the HyperparameterScheduler compatible with the CaseOptimizer
### Feature motivation
The HyperparameterScheduler is not compatible with the CaseOptimizer since the hyperparameters are attributes of the optimizers inside the CaseOptimizer. 

### Feature description
I propose one of the two possible solutions: 
Either we could give HyperparameterScheduler the optimizer as an argument. It could be called via ```  HyperparameterScheduler(schedule, hyperparameter, optimizer, verbose=0) ``` and the right optimizer inside the CaseOptimizer can be addressed. (@koenhelwegen) 

My second proposal would be to search the CaseOptimizer for optimizers that have the hyperparameter as attribute. Then the schedule can be applied to this optimizer only. The downside of this would be that in case there are two optimizers inside the CaseOptimizer that have a hyperparameter with the same name the schedule would be applied to both of them. I do not think this would happen very often but it could definitively be an issue. See code below for my second proposal. 
### Feature implementation
``` python 
class HyperparameterScheduler(tf.keras.callbacks.Callback):
    """Generic hyperparameter scheduler.
    # Arguments
    schedule: a function that takes an epoch index as input
        (integer, indexed from 0) and returns a new hyperparameter as output.
    hyperparameter: str. the name of the hyperparameter to be scheduled.
    verbose: int. 0: quiet, 1: update messages.
    """

    def __init__(self, schedule, hyperparameter, verbose=0):
        super(HyperparameterScheduler, self).__init__()
        self.schedule = schedule
        self.hyperparameter = hyperparameter
        self.verbose = verbose

    def on_epoch_begin(self, epoch, logs=None):
        for op in self.model.optimizer.optimizers:
            if hasattr(op, self.hyperparameter):

                hp = getattr(op, self.hyperparameter)
                try:  # new API
                    hyperparameter_val = tf.keras.backend.get_value(hp)
                    hyperparameter_val = self.schedule(epoch, hyperparameter_val)
                except TypeError:  # Support for old API for backward compatibility
                    hyperparameter_val = self.schedule(epoch)

                tf.keras.backend.set_value(hp, hyperparameter_val)

                if self.verbose > 0:
                    print(
                        f"Epoch {epoch + 1}: {self.hyperparameter} changning to {tf.keras.backend.get_value(hp)}."
                    )

    def on_epoch_end(self, epoch, logs=None):
        logs = logs or {}
        for op in self.model.optimizer.optimizers:
            if hasattr(op, self.hyperparameter):
                hp = getattr(op, self.hyperparameter)
                logs[self.hyperparameter] = tf.keras.backend.get_value(hp)
```

MariaHeuss:
This makes the HyperparameterScheduler compatible with the CaseOptimizer. The optimizer now needs to be passed as an argument as discussed in #347 

Fixes #347

### DIFF
--- a/larq/callbacks.py
+++ b/larq/callbacks.py
@@ -29,10 +29,7 @@ class HyperparameterScheduler(tf.keras.callbacks.Callback):
 
     def __init__(self, schedule, hyperparameter, optimizer=None, verbose=0):
         super(HyperparameterScheduler, self).__init__()
-        if optimizer is None:
-            self.optimizer = self.model.optimizer
-        else:
-            self.optimizer = optimizer
+        self.optimizer = optimizer if optimizer else self.model.optimizer
         self.schedule = schedule
         self.hyperparameter = hyperparameter
         self.verbose = verbose
--- a/larq/callbacks.py
+++ b/larq/callbacks.py
@@ -27,9 +27,12 @@ class HyperparameterScheduler(tf.keras.callbacks.Callback):
     verbose: int. 0: quiet, 1: update messages.
     """
 
-    def __init__(self, optimizer, schedule, hyperparameter, verbose=0):
+    def __init__(self, schedule, hyperparameter,optimizer=None, verbose=0):
         super(HyperparameterScheduler, self).__init__()
-        self.optimizer = optimizer
+        if optimizer is None:
+            self.optimizer = self.model.optimizer
+        else:
+            self.optimizer = optimizer
         self.schedule = schedule
         self.hyperparameter = hyperparameter
         self.verbose = verbose
--- a/larq/callbacks.py
+++ b/larq/callbacks.py
@@ -13,7 +13,11 @@ class HyperparameterScheduler(tf.keras.callbacks.Callback):
             ),
             default_optimizer=tf.keras.optimizers.Adam(0.01),
         )
-        callbacks = [HyperparameterScheduler(lambda x:0.001*(0.1**(x//30)), 'gamma', optimizer.optimizers[0])]
+        callbacks = [
+            HyperparameterScheduler(
+                lambda x:0.001*(0.1**(x//30)), 'gamma', optimizer.optimizers[0]
+            )
+        ]
         ```
     # Arguments
     optimizer: the optimizer that contains the hyperparameter that will be scheduled.
--- a/larq/callbacks.py
+++ b/larq/callbacks.py
@@ -4,26 +4,39 @@ import tensorflow as tf
 class HyperparameterScheduler(tf.keras.callbacks.Callback):
     """Generic hyperparameter scheduler.
 
+    !!! example
+        ```python
+        optimizer = lq.optimizers.CaseOptimizer(
+            (
+                lq.optimizers.Bop.is_binary_variable,
+                lq.optimizers.Bop(threshold=1e-6, gamma=1e-3),
+            ),
+            default_optimizer=tf.keras.optimizers.Adam(0.01),
+        )
+        callbacks = [HyperparameterScheduler(lambda x:0.001*(0.1**(x//30)), 'gamma', optimizer.optimizers[0])]
+        ```
     # Arguments
+    optimizer: the optimizer that contains the hyperparameter that will be scheduled.
     schedule: a function that takes an epoch index as input
         (integer, indexed from 0) and returns a new hyperparameter as output.
     hyperparameter: str. the name of the hyperparameter to be scheduled.
     verbose: int. 0: quiet, 1: update messages.
     """
 
-    def __init__(self, schedule, hyperparameter, verbose=0):
+    def __init__(self, optimizer, schedule, hyperparameter, verbose=0):
         super(HyperparameterScheduler, self).__init__()
+        self.optimizer = optimizer
         self.schedule = schedule
         self.hyperparameter = hyperparameter
         self.verbose = verbose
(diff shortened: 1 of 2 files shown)


leonoverweel:
Nice change, this'll be useful. Can you Black-format the example code and add a unit test?

larq/callbacks.py
```diff
@@ -4,21 +4,38 @@
...
+            default_optimizer=tf.keras.optimizers.Adam(0.01),
+        )
+        callbacks = [
+            HyperparameterScheduler(
+                lambda x:0.001*(0.1**(x//30)), 'gamma', optimizer.optimizers[0]
+            )
+        ]
+        ```
     # Arguments
+    optimizer: the optimizer that contains the hyperparameter that will be scheduled.
```
lgeiger: I think this should be optional (if not a `CaseOptimizer`), since most of the time you'd like to schedule parameters from `model.optimizer`. And this can lead to tricky bugs where you schedule an hyperparameter from an optimizer that is not actually used in your model.

MariaHeuss: Alright, I agree that this can lead to very hard to debug bugs (I had one of those myself) so I guess making this optional will both give an example how to use it properly and avoid bugs in the already existing code. I will make it optional.

MariaHeuss:
Will write the test and make the optimizer an optional argument.

larq/callbacks.py
MariaHeuss: I made it optional but did not yet add a test for the case that no optimizer argument is given. Do you think it would be good to have one?

MariaHeuss:
There is an error for black formatting, but as far as I am aware of my code is black formatted and from the error message it is not really clear to me what went wrong...

larq/callbacks.py
```diff
@@ -4,26 +4,46 @@
...
     verbose: int. 0: quiet, 1: update messages.
     """
 
-    def __init__(self, schedule, hyperparameter, verbose=0):
+    def __init__(self, schedule, hyperparameter, optimizer=None, verbose=0):
         super(HyperparameterScheduler, self).__init__()
+        if optimizer is None:
+            self.optimizer = self.model.optimizer
+        else:
+            self.optimizer = optimizer
```
leonoverweel: Small suggestion for conciseness:

```python
self.optimizer = optimizer if optimizer else self.model.optimizer
```

larq/callbacks.py
```diff
@@ -4,26 +4,43 @@
...
+                lq.optimizers.Bop(threshold=1e-6, gamma=1e-3),
+            ),
+            default_optimizer=tf.keras.optimizers.Adam(0.01),
+        )
+        callbacks = [
+            HyperparameterScheduler(
+                lambda x:0.001*(0.1**(x//30)), 'gamma', optimizer.optimizers[0]
+            )
+        ]
+        ```
```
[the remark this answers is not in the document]
MariaHeuss: Yes, this looks better. I changed it.

larq/callbacks_test.py
```diff
@@ -18,4 +24,76 @@ def on_epoch_end(self, epoch, logs={}):
...
+            optimizer=case_optimizer,
+            metrics=["accuracy"],
+        )
+
+        def scheduler(x):
+            return 1.0 / (1.0 + x)
+
+        cbk_gamma_scheduler = HyperparameterScheduler(
+            schedule=scheduler,
+            optimizer=model.optimizer.optimizers[0],
```
lgeiger:
```suggestion
            optimizer=bop,
```

larq/callbacks_test.py
```diff
@@ -18,4 +24,76 @@ def on_epoch_end(self, epoch, logs={}):
...
+            hyperparameter="gamma",
+            verbose=1,
+        )
+        cbk_threshold_scheduler = HyperparameterScheduler(
+            schedule=scheduler,
+            optimizer=model.optimizer.optimizers[0],
```
lgeiger:
```suggestion
            optimizer=bop,
```

larq/callbacks_test.py
```diff
@@ -18,4 +24,76 @@ def on_epoch_end(self, epoch, logs={}):
...
+            hyperparameter="threshold",
+            verbose=1,
+        )
+        cbk_lr_scheduler = HyperparameterScheduler(
+            schedule=scheduler,
+            optimizer=model.optimizer.optimizers[1],
```
lgeiger:
```suggestion
            optimizer=adam,
```

MariaHeuss:
> Cool, last minor comments

I would like to leave it like it is. If you have to reload the model and you define the callback like this it might create a second optimizer object with the callback but the actual optimizer will not be affected by the schedule. It is not relevant for this example but since people might take this as an orientation of how to use it I think this is the better way. What do you think?

lgeiger:
> I would like to leave it like it is. If you have to reload the model and you define the callback like this it might create a second optimizer object with the callback but the actual optimizer will not be affected by the schedule. It is not relevant for this example but since people might take this as an orientation of how to use it I think this is the better way. What do you think?

Ah I see what you mean, I agree it is not relevant here, i was just afraid that by using `model.optimizer.optimizers[0]` (which is specific to the case optimizer) you might pick the wrong one which might lead to annoying bugs.

PR #357, opened 2019-12-16
Simplify assign operations in Bop

lgeiger:
The use of `tf.compat.v1.assign` was only to make sure we execute the same code path as TF. I am not sure if this is necessary for TF 1.x.

Let's see what CI thinks.

### DIFF
--- a/larq/optimizers.py
+++ b/larq/optimizers.py
@@ -261,11 +261,9 @@ class Bop(tf.keras.optimizers.Optimizer):
         threshold = self._get_decayed_hyper("threshold", var_dtype)
         m = self.get_slot(var, "m")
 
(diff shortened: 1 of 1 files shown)


larq/optimizers.py
```diff
@@ -261,11 +261,9 @@ def _resource_apply_dense(self, grad, var):
         threshold = self._get_decayed_hyper("threshold", var_dtype)
         m = self.get_slot(var, "m")
 
-        m_t = tf.compat.v1.assign(
-            m, (1 - gamma) * m + gamma * grad, use_locking=self._use_locking
-        )
+        m_t = m.assign_add(gamma * (grad - m))
```
lgeiger: This algorithmic change might even speed up calculations very slightly.

PR #358, opened 2019-12-17
Use metaclass to allow direct instantiation of QuantizedVariable

lgeiger:
This allows us to remove the `create_quantized_variable` function at the expense of a slightly more complicated implementation. Users can now directly instantiate `QantizedVariable` and don't have to worry about correctly handling distribution strategies.

### DIFF
--- a/larq/quantized_variable.py
+++ b/larq/quantized_variable.py
@@ -7,7 +7,46 @@ from tensorflow.python.ops import resource_variable_ops
 from larq import quantized_scope
 
 
-class QuantizedVariable(tf.Variable):
+class QuantizedVariableFactory(tf.Variable.__class__):
+    """Factory to allow construction of QuantizedVariable to be overridden."""
+
+    def __call__(cls, variable, quantizer=None, precision=None):
+        """Creates a QuantizedVariable that wraps another variable.
+
+        This typically just returns `QuantizedVariable(variable)`. But, if the variable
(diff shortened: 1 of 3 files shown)


larq/quantized_variable.py
```diff
@@ -7,7 +7,46 @@
 from larq import quantized_scope
 
 
-class QuantizedVariable(tf.Variable):
+class QuantizedVariableFactory(tf.Variable.__class__):
```
lgeiger: `tf.Variable` is a `metaclass` itself, so we need to inherit from `tf.Variable.__class__` instead of `type` to prevent `metaclass` conflicts.

larq/quantized_variable.py
```diff
@@ -7,7 +7,46 @@
...
+            If `None`, `quantizer.precision` is used.
+
+        # Returns
+        A QuantizedVariable that wraps the variable.
+        """
+
+        if not isinstance(variable, DistributedVariable):  # type: ignore
+            return super().__call__(variable, quantizer, precision)
+
+        class QuantizedDistributedVariable(QuantizedVariable, variable.__class__):
```
lgeiger: Apparently this only works for TensorFlow 2.1 and will throw a metaclass conflict otherwise 😞

leonoverweel: So should we just do a factory instead?

PR #359, opened 2019-12-17
Use from_variable factory method to instantiate QuantizedVariables

lgeiger:
This allows creation of `QuantizedVariable` and `QuantizedDistributedVariable` using `QuantizedVariable.from_variable` following the naming conventions from `tf.data.Dataset.from_{tensors, tensor_slices, generator}`.

Closes #358

### DIFF
--- a/larq/layers_base.py
+++ b/larq/layers_base.py
@@ -3,7 +3,8 @@ import logging
 import tensorflow as tf
 
 from larq import metrics as lq_metrics
-from larq import quantized_scope, quantized_variable, quantizers
+from larq import quantized_scope, quantizers
+from larq.quantized_variable import QuantizedVariable
 
 log = logging.getLogger(__name__)
 
@@ -27,7 +28,7 @@ class BaseLayer(tf.keras.layers.Layer):
         # Wrap `getter` with a version that returns a `QuantizedVariable`.
         def getter(*args, **kwargs):
             variable = old_getter(*args, **kwargs)
-            return quantized_variable.create_quantized_variable(variable, quantizer)
+            return QuantizedVariable.from_variable(variable, quantizer)
 
         return super()._add_variable_with_custom_getter(name, getter=getter, **kwargs)
 
--- a/larq/layers_base.py
+++ b/larq/layers_base.py
@@ -2,8 +2,7 @@ import logging
 
 import tensorflow as tf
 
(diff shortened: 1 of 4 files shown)


larq/quantized_variable.py
```diff
@@ -39,6 +39,63 @@ def __init__(self, variable, quantizer=None, precision=None):
         self.quantizer = quantizer
         self.precision = precision or getattr(quantizer, "precision", None)
 
+    @classmethod
```
lgeiger: We could also use a `staticmethod` here, not sure what's better.

larq/layers_base.py
```diff
@@ -3,7 +3,8 @@
 import tensorflow as tf
 
 from larq import metrics as lq_metrics
-from larq import quantized_scope, quantized_variable, quantizers
+from larq import quantized_scope, quantizers
```
leonoverweel: I think it'd be cleaner to combine these `from larq` imports:

```python
from larq import quantized_scope, quantizers, metrics as lq_metrics
```

larq/quantized_variable.py
leonoverweel: The [`tf.data.Dataset.from_tensor`](https://github.com/tensorflow/tensorflow/blob/r2.0/tensorflow/python/data/ops/dataset_ops.py#L397) and [`tf.data.Dataset.from_tensor_slices`](https://github.com/tensorflow/tensorflow/blob/r2.0/tensorflow/python/data/ops/dataset_ops.py#L417) factories both use `staticmethod`, but reading [this StackOverflow answer](https://stackoverflow.com/a/15031999/1814949) and [this Software Engineering StackExchange answer](https://softwareengineering.stackexchange.com/a/166715), I think that you're right in using a `classmethod`. Maybe we should actually PR those TF Datasets methods to turn them into `classmethod`. 😄

larq/layers_base.py
lgeiger: I agree, but `isort` doesn't 😉

lgeiger: Changed the config in 9381fe27fbfb9ba79b3c0e8187401513944ef695

larq/quantized_variable.py
lgeiger: Cool, thanks for checking :+1:

PR #364, opened 2019-12-19
Use integer as kernel_quantizer to mark layers that should use Bop

### LINKED ISSUE #306: RFC: Quantized Variables

lgeiger:
As discussed in https://github.com/larq/larq/issues/306#issuecomment-553978646, this allows to use integers as `kernel_quantizers` in layers to mark them to be used by external optimizers like Bop.

This will be a breaking change because previously `Bop` was applied to **all** quantized layers which made it impossible to use `larq` layers with different precisions in a network. Now `kernel_quantizers=1` needs to be set on all layers that should be optimized by `Bop`. `kernel_quantizers="ste_sign"` keeps working too, but will raise a warning message.

Please review carefully since this is a user level API change.
Do you think it would useful to add a `Bop.legacy_is_binary_variable` method that exposes the deprecated old behaviour to ease the migration?

Closes #306

### DIFF
--- a/larq/layers.py
+++ b/larq/layers.py
@@ -69,7 +69,7 @@ class QuantDense(QuantizerBase, tf.keras.layers.Dense):
     input_quantizer: Quantization function applied to the input of the layer.
     kernel_quantizer: Quantization function applied to the `kernel` weights matrix
         or integer defining the desired precision if quantization should be handled
-        by an external optimizer.
+        by an specialized optimizer.
     kernel_initializer: Initializer for the `kernel` weights matrix.
     bias_initializer: Initializer for the bias vector.
     kernel_regularizer: Regularizer function applied to the `kernel` weights matrix.
@@ -166,7 +166,7 @@ class QuantConv1D(QuantizerBase, tf.keras.layers.Conv1D):
     input_quantizer: Quantization function applied to the input of the layer.
     kernel_quantizer: Quantization function applied to the `kernel` weights matrix
         or integer defining the desired precision if quantization should be handled
-        by an external optimizer.
+        by an specialized optimizer.
     kernel_initializer: Initializer for the `kernel` weights matrix.
     bias_initializer: Initializer for the bias vector.
     kernel_regularizer: Regularizer function applied to the `kernel` weights matrix.
@@ -276,7 +276,7 @@ class QuantConv2D(QuantizerBase, tf.keras.layers.Conv2D):
     input_quantizer: Quantization function applied to the input of the layer.
     kernel_quantizer: Quantization function applied to the `kernel` weights matrix
         or integer defining the desired precision if quantization should be handled
-        by an external optimizer.
+        by an specialized optimizer.
     kernel_initializer: Initializer for the `kernel` weights matrix.
     bias_initializer: Initializer for the bias vector.
     kernel_regularizer: Regularizer function applied to the `kernel` weights matrix.
@@ -393,7 +393,7 @@ class QuantConv3D(QuantizerBase, tf.keras.layers.Conv3D):
     input_quantizer: Quantization function applied to the input of the layer.
     kernel_quantizer: Quantization function applied to the `kernel` weights matrix
         or integer defining the desired precision if quantization should be handled
-        by an external optimizer.
+        by an specialized optimizer.
     kernel_initializer: Initializer for the `kernel` weights matrix.
     bias_initializer: Initializer for the bias vector.
     kernel_regularizer: Regularizer function applied to the `kernel` weights matrix.
@@ -503,7 +503,7 @@ class QuantDepthwiseConv2D(QuantizerDepthwiseBase, tf.keras.layers.DepthwiseConv
     input_quantizer: Quantization function applied to the input of the layer.
     depthwise_quantizer: Quantization function applied to the `depthwise_kernel`
         weights matrix or integer defining the desired precision if quantization should
-        be handled by an external optimizer.
+        be handled by an specialized optimizer.
     depthwise_initializer: Initializer for the depthwise kernel matrix.
     bias_initializer: Initializer for the bias vector.
     depthwise_regularizer: Regularizer function applied to the depthwise kernel matrix.
@@ -608,10 +608,10 @@ class QuantSeparableConv1D(QuantizerSeparableBase, tf.keras.layers.SeparableConv
     input_quantizer: Quantization function applied to the input of the layer.
     depthwise_quantizer: Quantization function applied to the depthwise kernel or
         nteger defining the desired precision if quantization should be handled by an
-        external optimizer.
+        specialized optimizer.
     pointwise_quantizer: Quantization function applied to the pointwise kernel or
         integer defining the desired precision if quantization should be handled by an
-        external optimizer.
+        specialized optimizer.
     depthwise_initializer: An initializer for the depthwise convolution kernel.
     pointwise_initializer: An initializer for the pointwise convolution kernel.
     bias_initializer: An initializer for the bias vector. If None, the default
@@ -741,10 +741,10 @@ class QuantSeparableConv2D(QuantizerSeparableBase, tf.keras.layers.SeparableConv
     input_quantizer: Quantization function applied to the input of the layer.
     depthwise_quantizer: Quantization function applied to the depthwise kernel matrix
         or integer defining the desired precision if quantization should be handled by
-        an external optimizer.
+        an specialized optimizer.
     pointwise_quantizer: Quantization function applied to the pointwise kernel matrix
         or integer defining the desired precision if quantization should be handled by
-        an external optimizer.
+        an specialized optimizer.
     depthwise_initializer: Initializer for the depthwise kernel matrix.
     pointwise_initializer: Initializer for the pointwise kernel matrix.
     bias_initializer: Initializer for the bias vector.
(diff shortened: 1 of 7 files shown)


larq/optimizers.py
```diff
@@ -205,6 +206,10 @@ class Bop(tf.keras.optimizers.Optimizer):
     Bop is a latent-free optimizer for Binarized Neural Networks (BNNs) and
     Binary Weight Networks (BWN).
 
+    To use Bop to train a model set `kernel_quantizer=1` for the layers where
+    binarization should be handled by Bop or use custom predicate function in the
+    `CaseOptimizer`.
```
leonoverweel: Suggestion:

To use Bop to train a model, set `kernel_quantizer=1` for the layers where training and binarization should be handled by Bop. This is the easiest, recommended approach. Alternatively, you can define a custom predicate function for the [`CaseOptimizer`](https://larq.dev/api/optimizers/#caseoptimizer) that captures the set of layers you want to assign to Bop (see Bop's [`is_binary_var()`](https://github.com/larq/larq/blob/master/larq/optimizers.py#L288) for example.)

larq/optimizers.py
```diff
@@ -286,11 +292,17 @@ def from_config(cls, config, custom_objects=None):
...
 
         # Arguments
         var: a `tf.Variable`.
         """
-        return "/kernel" in var.name and "quant_" in var.name
+        precision = getattr(var, "precision", None)
+        if precision == 1 and getattr(var, "quantizer", None) is not None:
+            warnings.warn(
+                f"Calling Bop with `quantizer={var.quantizer}`; this might result in "
+                "unexpected behaviour or won't do anything."
```
leonoverweel: Suggested addition:

To assign a layer to Bop, set `kernel_quantizer=1`.

larq/quantizers_test.py
```diff
@@ -56,7 +56,7 @@ def test_serialization_cls(self, ref_fn):
 
     def test_invalid_usage(self):
         with pytest.raises(ValueError):
-            lq.quantizers.get(42)
+            lq.quantizers.get(42.0)
```
leonoverweel: Hehe

larq/layers.py
```diff
@@ -67,7 +67,9 @@ class QuantDense(QuantizerBase, tf.keras.layers.Dense):
         no activation is applied (`a(x) = x`).
     use_bias: Boolean, whether the layer uses a bias vector.
     input_quantizer: Quantization function applied to the input of the layer.
-    kernel_quantizer: Quantization function applied to the `kernel` weights matrix.
+    kernel_quantizer: Quantization function applied to the `kernel` weights matrix
+        or integer defining the desired precision if quantization should be handled
+        by an external optimizer.
```
koenhelwegen: I would use "specialized optimizer" instead of "external optimizer"

lgeiger:
>  Such a user may simply set `kernel_quantizer=1`, throw in Adam, and expect it to result in binary weights, which if I understand correctly now would not be the case.

That's correct. I agree that the option of passing a desired precision as quantizer isn't the best from an API perspective. Though in your example I am afraid that we cannot avoid this problem in all cases (i.e. `kernel_quantizer=None` or  `kernel_quantizer=lambda x: 2*x` also don't result in quantized weights).

> What I would propose as a solution is to modify the behaviour of our QuantizedVariable class such that if the precision is set to 1 (or any other integer), the variable will always be quantized on the forward pass. As default quantizer I would propose the STE (or stepwise functions between -1 and 1 with clipped identity as backward pass for higher precisions).

I am against this solution due to two main issues: 1) Since this would introduces unnecessary operations in the graph it might degrade training performance. 2) Implicitly quantizing might not be what a users is trying to achieve and is also something that is not clearly visible via the API. E.g. You could think of a scaling "quantizer" `lambda x: alpha*x` with a trainable variable `alpha` where `x` is updated via Bop and `alpha` via a normal optimizer or a moving average (sorry for the contrived example). Things like that would not be possible to implement since `larq` would still quantize it to -1, +1.

How would you feel about a `NoOpQuantizer` (probably needs a better name), that is used instead of an integer? It could to some validation using `tf.assert` (we would need to check what the performance implications here are, but I don't think it would be noticeable). Then we can think about using the `kernel_quantizer=int` as a shorthand to that pseudo quantizer much like the string representations of `ste_sign`. We could enable validation by default, but through the `NoOpQuantizer` class advanced users would be able to turn it off and have full flexibility.

koenhelwegen:
I'm not too worried about additional operations - I think the performance degradation will be negligible and would prefer (more) robust behaviour (I realize we cannot cover all cases, but introducing the possibility of a vacuous `kernel_quantizer=1` seems like a bad idea).

> Implicitly quantizing might not be what a users is trying to achieve and is also something that is not clearly visible via the API. E.g. You could think of a scaling "quantizer" lambda x: alpha*x with a trainable variable alpha where x is updated via Bop and alpha via a normal optimizer or a moving average (sorry for the contrived example). Things like that would not be possible to implement since larq would still quantize it to -1, +1.

I'm not sure if I understand you correctly. The user can still set a quantizer and overwrite the default. It is just that in a quantized execution scope, a QuantizedVariable object with precision 1 should never return a bunch of floating points.

> How would you feel about a NoOpQuantizer (probably needs a better name), that is used instead of an integer? It could to some validation using tf.assert (we would need to check what the performance implications here are, but I don't think it would be noticeable). Then we can think about using the kernel_quantizer=int as a shorthand to that pseudo quantizer much like the string representations of ste_sign. We could enable validation by default, but through the NoOpQuantizer class advanced users would be able to turn it off and have full flexibility.

I would rather avoid this; i think `kernel_quantizer=1` is already stretching the meaning of 'quantizer'. If we want to use it just a marker for bop, I would prefer adding a `precision` argument to the layer.

lgeiger:
> I'm not too worried about additional operations - I think the performance degradation will be negligible

I can get some rough performance numbers next week, but unless we have tested this claim, I'd be against adding implicit behaviour there.

@koenhelwegen Since you raise concerns about the overall API design, I am closing this PR for now. We discussed possible options in https://github.com/larq/larq/issues/306#issuecomment-553978646 and agreed on this (or the `NoOpQuantizer` as the best options). Please add your proposed API design there so we can revisit this discussion and go from there.

PR #381, opened 2020-01-14
Fix bug in HyperParameterScheduler

jneeven:
It would previously crash when used when not specifying an optimizer, since `self.model` does not yet exist in ` __init__`. `set_model` is called once as part of `model.fit`, so this seems the best place for it.


jneeven:
I accidentally removed the `LogHistory` from the `callbacks_test`, but it wasn't used anyway. I discussed with @lgeiger and @leonoverweel  and we concluded it should be fine to remove it for now.

larq/callbacks.py
```diff
@@ -24,19 +27,32 @@ class HyperparameterScheduler(tf.keras.callbacks.Callback):
...
     """
 
-    def __init__(self, schedule, hyperparameter, optimizer=None, verbose=0):
+    def __init__(
+        self,
+        schedule: Callable,
+        hyperparameter: str,
+        optimizer: Optional[keras.optimizers.Optimizer] = None,
+        verbose: Optional[int] = 0,
+    ):
```
lgeiger: Thanks for the typehints. Unfortunately they don't show up nicely in our docs. I guess I need to refine our API docs generation. Though I think this shouldn't block this PR.

leonoverweel: @lgeiger looks like pydoc-markdown [doesn't have support for this yet](https://github.com/NiklasRosenstein/pydoc-markdown/issues/86).

lgeiger: We could also try out https://github.com/keras-team/keras-autodoc which looks similar.

lgeiger: For now we simply ignore the typehints in our rendered docs. I made some quick changes in to https://github.com/lgeiger/pydoc-markdown/commit/d6628ebd9f2ad5429c6898e63e5626873bdfc631

PR #385, opened 2020-01-15
Support per-step scheduling in HyperParameterScheduler

jneeven:
This is useful for e.g. warmup, where you want to change a hyperparameter at higher resolution than full epochs. I have tested that this works, but will write some explicit tests for this later.

Side note: I have also made some schedule classes and a `CombinedSchedule` class on a different branch, so that we can easily combine several schedules. This works nicely with the per-step `HyperParameterScheduler` and will probably be upstreamed later if this is desired.


larq/callbacks.py
```diff
@@ -40,6 +43,15 @@ def __init__(
...
         self.verbose = verbose
 
+        # TODO: may want to make this a boolean instead for efficient comparison,
+        # need to test impact on speed.
+        if unit == "epoch" or unit == "step":
+            self.unit = unit
+        else:
+            raise ValueError(
+                f"HyperparameterScheduler.unit can only be 'step' or 'epoch'. Received value '{unit}'"
+            )
```
leonoverweel: Suggestion:

```python
if not unit in ["epoch", "step"]:
    raise ValueError(...)
```

larq/callbacks.py
```diff
@@ -40,6 +43,15 @@ def __init__(
         self.hyperparameter = hyperparameter
         self.verbose = verbose
 
+        # TODO: may want to make this a boolean instead for efficient comparison,
+        # need to test impact on speed.
```
leonoverweel: Since this just gets called once in the `__init__()`, I don't think we need to worry about speed.

larq/callbacks.py
```diff
@@ -24,6 +24,8 @@ class HyperparameterScheduler(tf.keras.callbacks.Callback):
     schedule: a function that takes an epoch index as input
         (integer, indexed from 0) and returns a new hyperparameter as output.
     hyperparameter: str. the name of the hyperparameter to be scheduled.
+    unit: str (optional), what interval unit to change the hyperparameter at. Can be
+        either "epoch" (default) or "step".
```
leonoverweel: I think `interval` would be a better name for this parameter.

larq/callbacks.py
```diff
@@ -52,20 +64,32 @@ def set_model(self, model: keras.models.Model):
...
-            )
+    def on_epoch_begin(self, epoch: int, logs: Optional[Dict] = None):
+        if self.unit == "epoch":
+            hp = self.set_hyperparameter(epoch)
+
+            if self.verbose > 0:
+                print(
+                    f"Epoch {epoch + 1}: {self.hyperparameter} changing "
+                    + f"to {tf.keras.backend.get_value(hp)}."
+                )
```
leonoverweel: I don't think we're splitting strings for the 88-char limit, so you can keep these as one line.

larq/callbacks.py
```diff
@@ -52,20 +64,32 @@ def set_model(self, model: keras.models.Model):
...
+
+    def on_batch_begin(self, batch: int, logs: Optional[Dict] = None):
+        if self.unit == "step":
+            # We use optimizer.iterations (i.e. global step), since batch only
+            # reflects the batch index in the current epoch.
+            self.set_hyperparameter(self.optimizer.iterations)
```
leonoverweel: Maybe add an `if verbose: print(...)` here too, since it's also there for the epoch version.

larq/callbacks.py
leonoverweel: Also (maybe check this) I think epochs are reported 0-indexed everywhere, so I wouldn't do `epoch + 1` here.

jneeven: This code was already in the hyperparameter before, so I assume this is okay? I splitted the string just because I think it looks uglier if I don't

larq/callbacks.py
jneeven: It's not the init I'm worried about; it's the string comparisons that potentially happen each batch. Just added the comment here because this is the place where we'd need to create that bool. I'll test whether this actually makes a difference though :+1:

jneeven: I have run an experiment using a per-step schedule, and the time per epoch is actually faster than it was before (while the setup otherwise remains identical), so I think it's safe to assume that there is no significant slowdown (of course I also strongly doubt that this actually caused a speedup, whatever effect this has is just outweighed by inherent variability in GPU time)

larq/callbacks.py
leonoverweel: We don't split strings or f-strings anywhere else in the codebase as far as I can tell, so I think it's better to be consistent and also keep this string as one line.

larq/callbacks.py
lgeiger: The [TensorBoardCallback](https://keras.io/callbacks/#tensorboard) has an `update_freq` which can be either `'batch'` or `'epoch'` or integer. I think it would be good to align our selfs with a similar naming scheme.

larq/callbacks.py
```diff
@@ -52,20 +64,32 @@ def set_model(self, model: keras.models.Model):
                 f'Optimizer must have a "{self.hyperparameter}" attribute.'
             )
 
-    def on_epoch_begin(self, epoch: int, logs: Optional[Dict] = None):
+    def set_hyperparameter(self, t: int):
```
lgeiger: What about renaming `t` to `epoch_or_step`?

larq/callbacks.py
```diff
@@ -52,20 +64,32 @@ def set_model(self, model: keras.models.Model):
...
-                f"Epoch {epoch + 1}: {self.hyperparameter} changing to {tf.keras.backend.get_value(hp)}."
-            )
+    def on_epoch_begin(self, epoch: int, logs: Optional[Dict] = None):
+        if self.unit == "epoch":
+            hp = self.set_hyperparameter(epoch)
+
+            if self.verbose > 0:
+                print(
+                    f"Epoch {epoch + 1}: {self.hyperparameter} changing "
+                    + f"to {tf.keras.backend.get_value(hp)}."
```
lgeiger: If you want to split the line, I'd remove the `+`:
```suggestion
                    f"to {tf.keras.backend.get_value(hp)}."
```
I am also fine with keeping it in one line since we don't have a linter to enforce this limit. Whatever looks more readable in this case.

larq/callbacks.py
lgeiger: Nice, I didn't realize that we now have access to the optimizer iterations.

larq/callbacks.py
AdamHillier: I like splitting strings as well - Zookeeper splits all the strings - not that it really matters too much. But I believe if you do choose to split them you can get rid of the `+`.

larq/callbacks.py
jneeven: I don't like that solution, seems a hideous variable name. I get your point, though I don't see any better alternatives either.

jneeven: Awesome. I just made some final changes, should be good to merge now.

larq/callbacks.py
```diff
@@ -32,6 +34,7 @@ def __init__(
         schedule: Callable,
         hyperparameter: str,
         optimizer: Optional[keras.optimizers.Optimizer] = None,
+        update_freq: Optional[str] = "epoch",
```
lgeiger:
```suggestion
        update_freq: str = "epoch",
```

jneeven:
If this is approved, please do not immediately merge it. I will verify that the tensorboard logging still works (currently doesn't seem to be the case with object detection).

larq/callbacks.py
```diff
@@ -39,6 +42,14 @@ def __init__(
         self.hyperparameter = hyperparameter
         self.verbose = verbose
 
+        if update_freq not in ["epoch", "step"]:
```
lgeiger: I tend to prefer the following, but not strong opinion:
```suggestion
        if update_freq not in ("epoch", "step"):
```

larq/callbacks.py
```diff
@@ -51,6 +62,30 @@ def set_model(self, model: keras.models.Model) -> None:
...
+        keras.backend.set_value(hp, hyperparameter_val)
+        return hp
+
+    def on_batch_begin(
+        self, batch: int, logs: Optional[MutableMapping[str, Any]] = None
+    ) -> None:
+        if self.update_freq == "step":
+            # We use optimizer.iterations (i.e. global step), since batch only
+            # reflects the batch index in the current epoch.
+            hp = self.set_hyperparameter(self.optimizer.iterations.numpy())
```
lgeiger: This won't work for older versions of TensorFlow or when running in graph mode. I think the following might work:
```suggestion
            hp = self.set_hyperparameter(keras.backend.get_value(self.optimizer.iterations))
```
Or we can move the `keras.backend.get_value` inside `set_hyperparameter`

jneeven: This should work, thanks :+1:

larq/callbacks.py
jneeven: Why? I have the exact opposite view but I'm curious to hear your reasoning :sweat_smile:

lgeiger: Again, not a well motivated reason to change that. I think I had in mind that tuples are a bit smaller and have faster element lookup if doing `for x in sth`, but I might be totally wrong and it definitely doesn't matter here. I might be just more used to seeing a tuple version of this...
Feel free to leave it as is.

jneeven:
It passed :tada:

larq/callbacks.py
```diff
@@ -51,26 +64,50 @@ def set_model(self, model: keras.models.Model) -> None:
...
+        # We use optimizer.iterations (i.e. global step), since batch only
+        # reflects the batch index in the current epoch.
+        batch = keras.backend.get_value(self.optimizer.iterations)
+        hp = self.set_hyperparameter(batch)
+
+        if self.verbose > 0:
+            print(
+                f"Batch {batch}: {self.log_name} changing"
+                f" to {keras.backend.get_value(hp)}."
+            )
```
leonoverweel: This way it fits into one < 88 char line 🤓

```python
            print(
                f"Batch {batch}: {self.log_name} is now {keras.backend.get_value(hp)}."
            )
```

larq/callbacks.py
```diff
@@ -51,26 +64,50 @@ def set_model(self, model: keras.models.Model) -> None:
...
+            return
+
+        hp = self.set_hyperparameter(epoch)
 
         if self.verbose > 0:
             print(
-                f"Epoch {epoch + 1}: {self.hyperparameter} changing to {keras.backend.get_value(hp)}."
+                f"Epoch {epoch}: {self.log_name} changing"
+                f" to {keras.backend.get_value(hp)}."
             )
```
leonoverweel: This way it fits into one < 88 char line 🤓

```python
            print(
                f"Batch {batch}: {self.log_name} is now {keras.backend.get_value(hp)}."
            )
```

jneeven:
Should be good to go now. I also added a `log_name` argument; this is useful e.g. when you use the `CaseOptimizer` and want to change the same variable on both optimizers with a different schedule for each. You can then log them as e.g. `binary_lr` and `fp_lr` instead of having just one of them in tensorboard.

larq/callbacks.py
```diff
@@ -31,14 +33,25 @@ def __init__(
         schedule: Callable,
         hyperparameter: str,
         optimizer: Optional[keras.optimizers.Optimizer] = None,
+        update_freq: str = "epoch",
         verbose: int = 0,
+        log_name: Optional[str] = None,
```
leonoverweel: Can you add a docstring for this (also explaining when you'd pass it)?

jneeven: I thought I did

jneeven: I did, but on the wrong repo. Amazing :sweat_smile:

jneeven:
> tada, nice change!

This is only the beginning :smirk:

Issue #343, opened 2019-11-29
RFC: Quantizers as subclasses of keras.layer.Layer


lgeiger:
This Request for Comments (RFC) outlines a proposed change to API and internal implementation of `larq.quantizers`. It would be good to get feedback on this, so please ask questions if I need to elaborate more in some areas.

## Objective

We currently have three possible ways of accessing `larq.quantizers`:
1. As a string: `ste_sign`
2. As a class with optional configurable arguments: `SteSign(clip_value=1.0)`
3. As a function: `larq.quantizers.ste_sign`

This raises a few question as it can be confusing due to having different ways to achieve the same  thing which we also discussed in https://github.com/larq/larq/issues/246#issuecomment-536531445 previously.
Also the way we currently handle metrics is suboptimal since they are added to `larq.layers` directly which makes implementation messy and doesn't allow for metrics that access parts of the custom gradients. Which make it hard to implement ideas like #245 in a general way.

## Design Proposal

In this RFC I'd like to propose three changes to the current API and implementation which should solve the above problems.

### Proposal 1: Unify `larq.quantizers` API #387

The functional way (see option 3. above) of calling `larq.quantizers` doesn't add much value to `larq` and is only a shorthand for calling the class based API which has much more flexibility. This adds overhead in the source code of `larq`  and will make Proposal 2 and 3 unnecessarily hard to implement. It might also be confusing for people when looking at the API docs to see two almost identical ways to use `larq.quantizers`. I propose to remove this option.

### Proposal 2: `larq.quantizers` should subclass `keras.layers.Layer` #387

It is currently possible to use custom `keras.layers.Layer`s as `larq.quantizers` (see #322). This can be useful when implementing quantizers with trainable variables or other variables that keep a running mean of scaling, commonly used in 8-bit quantizers. This change is only an implementation detail needed for Proposal 3 and doesn't change the way people currently use `larq.quantizers`.

### Proposal 3: Add metrics to `larq.quantizers` instead of `larq.layers`

The way we currently add metrics like `flip_ratio` to `larq.layers` is messy and doesn't allow for metrics accessing custom gradients which blocks #245.
After implementing proposal 2, we can add relevant metrics directly to `larq.quantizers` which would allow for better control over which metrics to add to which quantizer and allows direct access to custom gradients.

## Implementation

The implementation could looks something like this (not including configuration of metrics or special handling for eval/train):
```python
import tensorflow as tf
import larq as lq


class GradientFlow(tf.keras.metrics.Metric):
[...]

PR #387, opened 2020-01-16
RFC: Quantizers as subclasses of keras.layer.Layer

### LINKED ISSUE #246: Documentation (Regarding to JOSS review)
Some suggestions:
In the quantizers documentation:
1. It seems that SteSign/ste_sign (if they are the same) is repeated twice [here](https://larq.dev/api/quantizers/#ste_sign). The same for swish_sign/SwishSign.
2. These quantizers (magnitude_aware_sign,swish_sign)have no formula, nor description.

lgeiger:
This PR implements Proposal 1 and 2 of #343. The diff is quite large so it makes sense to carefully review this commit by commit.

Closes #246

### DIFF
--- a/docs/guides/bnn-architecture.md
+++ b/docs/guides/bnn-architecture.md
@@ -19,8 +19,8 @@ import larq as lq
 ...
 x_out = lq.layers.QuantDense(
     units=1024,
-    input_quantizer=lq.quantizers.ste_sign,
-    kernel_quantizer=lq.quantizers.ste_sign,
+    input_quantizer=lq.quantizers.SteSign(clip_value=1.0),
+    kernel_quantizer=lq.quantizers.SteSign(clip_value=1.0),
     kernel_constraint=lq.constraints.WeightClip(clip_value=1.0),
     )(x_in)
 ```
@@ -77,8 +77,8 @@ def conv_with_shortcut(x):
         filters=filters,
         kernel_size=3,
         padding="same",
-        input_quantizer=lq.quantizers.ste_sign,
-        kernel_quantizer=lq.quantizers.ste_sign,
+        input_quantizer=lq.quantizers.SteSign(clip_value=1.0),
+        kernel_quantizer=lq.quantizers.SteSign(clip_value=1.0),
         kernel_initializer="glorot_normal",
         kernel_constraint=lq.constraints.WeightClip(clip_value=1.0),
         use_bias=False,
@@ -105,8 +105,8 @@ x = lq.layers.QuantConv2D(
     filters=512,
     kernel_size=3,
     padding="same",
-    input_quantizer=lq.quantizers.ste_sign,
-    kernel_quantizer=lq.quantizers.ste_sign,
+    input_quantizer=lq.quantizers.SteSign(clip_value=1.0),
+    kernel_quantizer=lq.quantizers.SteSign(clip_value=1.0),
     kernel_constraint=lq.constraints.WeightClip(clip_value=1.0),
     use_bias=False,
     )(x)
--- a/docs/guides/key-concepts.md
+++ b/docs/guides/key-concepts.md
@@ -3,7 +3,7 @@ Below, we summarize the key concepts you need to understand to work with BNNs.
 
 ## Quantizer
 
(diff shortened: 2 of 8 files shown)


larq/quantizers.py
```diff
@@ -153,70 +83,8 @@ def _scaled_sign(x):  # pragma: no cover
     return 1.3 * ste_sign(x)
 
 
-@utils.set_precision(1)
-def magnitude_aware_sign(x, clip_value=1.0):
```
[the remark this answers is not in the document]
jneeven: I see that you eventually removed all of them, so I guess it doesn't matter

jneeven:
I think this is overall a very nice change, perhaps it would be nice to have a `QuantizerLayer` superclass that we can add metrics to?

lgeiger:
> Whats up with ApproxSign? Was it not in here before? (referring to your first commit)

Yes I think it wasn't here because it doesn't have any configurable parameters like `clip_value` (or maybe we just forgot).

> I think this is overall a very nice change, perhaps it would be nice to have a `QuantizerLayer` superclass that we can add metrics to?

Yes I think that would be the right thing to do. I am not sure yet if it should be a public API users could use to write their own quantizers, of if it should just be an implementation detail. I still need to check why tests are failing for TF != 2.1.0 before I'll start cleaning this PR up.

docs/guides/key-concepts.md
```diff
@@ -3,7 +3,7 @@ Below, we summarize the key concepts you need to understand to work with BNNs.
 
 ## Quantizer
 
-The transformation from high-precision Neural Networks to Quantized Neural Networks (QNNs) is achieved by [quantization](https://en.wikipedia.org/wiki/Quantization_(signal_processing)).
+The transformation from high-precision Neural Networks to Quantized Neural Networks (QNNs) is achieved by [quantization](<https://en.wikipedia.org/wiki/Quantization_(signal_processing)>).
```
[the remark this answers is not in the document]
lgeiger: :+1: I think that was [Prettier](https://prettier.io/) just being confused.

lgeiger:
This should now be ready for a thorough review

docs/guides/bnn-architecture.md
```diff
@@ -19,8 +19,8 @@ import larq as lq
 ...
 x_out = lq.layers.QuantDense(
     units=1024,
-    input_quantizer=lq.quantizers.ste_sign,
-    kernel_quantizer=lq.quantizers.ste_sign,
+    input_quantizer=lq.quantizers.SteSign(clip_value=1.0),
+    kernel_quantizer=lq.quantizers.SteSign(clip_value=1.0),
     kernel_constraint=lq.constraints.WeightClip(clip_value=1.0),
```
leonoverweel: It's nice that the `SteSign()` instantiations are now more similar to the `WeightClip()` ones.

larq/conftest.py
```diff
@@ -19,6 +19,7 @@ def graph_mode():
     with context.graph_mode():
         with tf.compat.v1.Session().as_default():
             yield
+            tf.keras.backend.clear_session()
```
leonoverweel: Why is this necessary?

larq/quantized_variable_test.py
```diff
@@ -21,7 +21,7 @@ def test_inheritance(distribute_scope):
     assert isinstance(quantized_variable, DistributedVariable) is distribute_scope  # type: ignore
 
 
-def test_read(distribute_scope, eager_and_graph_mode):
+def test_read(eager_and_graph_mode, distribute_scope):
```
leonoverweel: Why are these now all flipped?

larq/conftest.py
lgeiger: This is required for TensorFlow < 2.1 see f43aa7b25495a6ede2f9139f10559dfd8e450713

larq/quantized_variable_test.py
lgeiger: Due to the change above, we need to set the `graph_mode` fixture before setting a distribution scope, otherwise the cleanup will error due to the graph beeing clean up, before exiting the distribution scope.

Issue #309, opened 2019-11-07
RFC: Use type hints in larq source code


lgeiger:
## Objective

Python 3.6 and above supports [type hints](https://www.python.org/dev/peps/pep-0484/) and optional statically typing of Python code. Since its introduction is has seen a lot of adoption in many Python projects in particular in larger code bases. Type hints are completely optional and are stripped by the Python parser before runtime.

I am not the only one @plumerai anymore who likes type hints for Python and we only support Python 3.6+ already so I'd like to discuss if we should introduce type hints to the `larq` code base. I have never used typings in a big project, but below is a list of some pros and cons I can think of on top of my head.

## Pros
- Makes it easy to reason about the code due to the self documenting nature of statically typed functions
- Helps with code review
- Can catch subtle bugs of in places of insufficient code coverage
- Has good integrations for editors that help with autocomplete and linting

## Cons
- TensorFlow doesn't use type hints so I don't know how well type stubs are maintained
- Needs additional setup on CI and to maintain type stubs
- Might be confusing for novice Python users since it adds new syntax to the `larq` code base

## Tooling
- [MyPy](http://www.mypy-lang.org/) (the default)
- [Typeshed](https://github.com/python/typeshed)
- [Pyre](https://pyre-check.org/)
- [Pytype](https://google.github.io/pytype/)

AdamHillier:
Yes.

leonoverweel:
Yes, this would be nice.

AdamHillier:
Since 2b262bfeded0afaa9f30088e43399863c88f1337 we do run `pytype` in CI but it doesn't do too much yet because there aren't many type hints around.

koenhelwegen:
Sounds like a great idea! One small remark: as TensorFlow currently doesn't have typehints I would prefer to keep all examples provided in the documentation free of typehints, so as to not raise the threshold for users who are not familiar with them.

Btw, I know people are working on making type hints more compatible with numpy arrays, do we know if similar efforts are underway for TF tensors?

AdamHillier:
> Sounds like a great idea! One small remark: as TensorFlow currently doesn't have typehints I would prefer to keep all examples provided in the documentation free of typehints, so as to not raise the threshold for users who are not familiar with them.

The most important thing is for the external API to have type hints so that this gets picked up in editors/IDEs; any private methods or tests don't need them, and examples definitely don't.

lgeiger:
Great, thanks for the comments. I set the status to `RFC - accepted`.

[...]

PR #391, opened 2020-01-19
(Partially) add type hints. See #309. 

AdamHillier:


larq/constraints.py
```diff
@@ -34,13 +36,13 @@ class WeightClip(tf.keras.constraints.Constraint):
...
-    def __init__(self, clip_value=1):
+    def __init__(self, clip_value: float = 1):
         self.clip_value = clip_value
 
-    def __call__(self, x):
+    def __call__(self, x: tf.Tensor) -> tf.Tensor:
         return tf.clip_by_value(x, -self.clip_value, self.clip_value)
 
-    def get_config(self):
+    def get_config(self) -> Dict:
```
lgeiger: General Python typehints question: Is using the builtin `dict` equivalent to `typings.Dict` in this case?
```suggestion
    def get_config(self) -> dict:
```

larq/models.py
```diff
@@ -415,9 +435,6 @@ def summary(model, print_fn=None, include_macs=True):
             "`input_shape` argument in the first layer(s) for automatic build."
         )
 
-    if print_fn is None:
-        print_fn = print
-
```
lgeiger: I think we still need this.

AdamHillier: I made it the default value of the arg above so I don't think so?

https://github.com/larq/larq/blob/246a64a4d0ce7f3044cafee58dbb04b42b94879b/larq/models.py#L395

lgeiger: Ahh sorry, missed that. I think there was a reason, why I didn't do that in the first place. But I can't remember and CI is happy with it :+1:

larq/constraints.py
AdamHillier: Not 100% sure about the difference, but now I look into it apparently it's better to use `typing.Mapping` for such things, so I will edit.

PR #392, opened 2020-01-19
Fix typo in `DoReFa` docs.

AdamHillier:


larq/quantizers.py
```diff
@@ -620,8 +620,8 @@ class DoReFaQuantizer(QuantizerFunctionWrapper):
     \\[
     q(x) = \begin{cases}
     0 & x < \frac{1}{2n} \\\
-    \frac{i}{n} & \frac{2i-1}{2n} < |x| < \frac{2i+1}{2n} \text{ for } i \in \\{1,n-1\\}\\\
-     1 & \frac{2n-1}{2n} < x
+    \frac{i}{n} & \frac{2i-1}{2n} <= x < \frac{2i+1}{2n} \text{ for } i \in \\{1,n-1\\}\\\
+     1 & \frac{2n-1}{2n} <= x
     \end{cases}
```
jamescook106: What equation in https://arxiv.org/pdf/1606.06160.pdf is this referring to?

AdamHillier: Good question - I wasn't basing this on the paper, just on the implementation https://github.com/larq/larq/blob/a6a05509bbf2ccc3130fc5d05f7a4bf10f0c2d20/larq/quantizers.py#L424-L431

AdamHillier: @MariaHeuss?

MariaHeuss: It is the quantize_k in equation (5). I think in the paper it was not specified in which direction to round and it also does not really matter. tf.round is rounding half to even so I guess just  not specifying '=' case would be best?

AdamHillier: Ah nice I had forgotten about the rounding behaviour. Okay, I'm happy to revert the `<=` but I think the previous `|x|` was incorrect right, should be `x`?

MariaHeuss: Yeah, you are right about this, nice that you spottet it!

PR #405, opened 2020-02-04
Add schedule classes

jneeven:
These are useful for combining several schedules, e.g. linear warmup and cosine decay, in a clean way that also ensures the hyperparameter is logged to tensorboard. The `Schedule` class accepts any callable, including keras schedules like `keras.optimizers.schedules.PolynomialDecay`, or a custom function. All of these have been tested on object detection and are very convenient there.

Using these with the `HyperparameterScheduler` is simple:
```python
     if self.learning_rate_schedule:
            callbacks.append(
                lq.callbacks.HyperparameterScheduler(
                    self.learning_rate_schedule,
                    "lr",
                    update_freq=self.learning_rate_schedule.update_freq,
                )
            )
```

Example for linear warmup + cosine decay:
```python
    @cached_property
    def main_schedule(self):
        main_epochs = self.epochs - self.warmup_duration
        # The -1 is necessary to actually reach 0.
        return utils.Schedule(
            schedule=keras.experimental.CosineDecay(
                self.initial_lr,
                (self.steps_per_epoch * main_epochs) - 1,
                name="main_schedule",
            )
        )

    @cached_property
    def learning_rate_schedule(self):
        warmup_schedule = utils.Schedule(
            schedule=keras.optimizers.schedules.PolynomialDecay(
                0.0, self.steps_per_epoch * self.warmup_duration, self.initial_lr,
            )
        )
        return utils.CombinedSchedule(
            schedules=[warmup_schedule, self.main_schedule],
            starting_times=[self.warmup_duration * self.steps_per_epoch],
        )
```

It's a draft because it needs more tests etc, but feel free to review the current code!

TODO:
- [ ] Write unit tests
- [ ] Add docstrings
- [ ] Add some code examples in docstrings

### DIFF
--- a/larq/schedules.py
+++ b/larq/schedules.py
@@ -1,0 +1,124 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+from typing import Any, Callable, List, Optional
+
+from cached_property import cached_property
+
+
+class AbstractSchedule(ABC):
+    # TODO: Make this a component once Zookeeper has a post-init hook.
+
+    update_freq: str = "step"
+    subtract_start_time: bool = True
+
+    @abstractmethod
+    def __call__(self, t: int, start_time: int = 0) -> float:
+        raise NotImplementedError
+
+
+@dataclass
+class StepDecaySchedule(AbstractSchedule):
+    # TODO: needs documentation
+
+    initial_value: float
+    decay_steps: List[int]
+    decay_factor: float
+    update_freq: str = "epoch"
+    subtract_start_time: bool = False
+
+    def __call__(self, t: int, start_time: int = 0) -> float:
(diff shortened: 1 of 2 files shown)


jneeven:
@lgeiger  any ideas on this? Do we want this in Larq or elsewhere?

lgeiger:
> @lgeiger any ideas on this? Do we want this in Larq or elsewhere?

I personally would prefer having it elsewhere, since it is not directly related to quantization. To keep the scope of larq focused.

jneeven:
> > @lgeiger any ideas on this? Do we want this in Larq or elsewhere?
> 
> I personally would prefer having it elsewhere, since it is not directly related to quantization. To keep the scope of larq focused.

Sounds good, shall I move it to research temp?

PR #418, opened 2020-02-17
Restructure docs and add compute-engine tab

lgeiger:
This PR restructures the docs and changes the CSS to align it with the future landing page.

Here is a preview of how it would look like: https://larq-git-docs-restructuring.plumerai-larq.now.sh/


AdamHillier:
Screen-Recording-2020-02-17-at-1

I know browser compatability is boring, and I have no idea if this was introduced in this commit, but on Firefox the logo gets removed and re-loaded on each click of the top nav, as demonstrated by this screen recording (the GIF converter  I used dropped several frames so it looks like the reloading didn't happen for the middle item, but it did).

This actually isn't just the top nav, it happens when clicking on anything in the left-hand nav bar too.

Works as expected on Chrome.

lgeiger:
@AdamHillier Just checked this issue exists https://larq.dev/ as well

AdamHillier:
> @AdamHillier Just checked this issue exists https://larq.dev/ as well

Ah true, okay. The Mozillian in me hates it... Almost certainly an issue in mkdocs/other library though. Will open an issue.

lgeiger:
Note, the links are currently broken, which I will fix now

PR #424, opened 2020-02-17
Workaround #422

### LINKED ISSUE #422: Minor layout bug with docs headers
When the header length is just under the width of the `md-content` \<div\>, there is incorrectly a double line break before the content. This can be seen below (the "Can I use Larq only ..." block):

image

The issue arises because of the "anchor" button, which is invisible unless you hover over it. When hovering, the header looks like this:

image

When not hovering over it, it overflows onto the next line but is invisible so it looks like an (additional) line break before the content.

lgeiger:
This makes the heading shorter to workaround #422

Closes #422


AdamHillier:
Oh boy, I guess that's *technically* a workaround...

Issue #410, opened 2020-02-07
Merge `QuantizerBase` and `BaseLayer`


leonoverweel:
### Feature motivation
We now have both a `BaseQuantizer` (in `quantizers.py`) and a `QuantizerBase` (in `layers_base.py`). This is confusing. Following #402, it should be easier to fuse `QuantizerBase` into `BaseLayer`. See [this PR comment](https://github.com/larq/larq/pull/402#discussion_r375925215).

### Feature description
Combine the `QuantizerBase` into `BaseLayer` classes.

PR #436, opened 2020-02-28
Cleanup base layers

lgeiger:
This is a first step towards #410

### DIFF
--- a/larq/layers_base.py
+++ b/larq/layers_base.py
@@ -32,11 +32,12 @@ class BaseLayer(tf.keras.layers.Layer):
             "input_quantizer": quantizers.serialize(self.input_quantizer),
         }
 
-    def get_quantizer(self, name) -> Optional[Quantizer]:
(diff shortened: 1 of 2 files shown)


lgeiger:
> (Why) do we need the separation between BaseLayer and QuantizerBase?

First there was `QuantizerBase`, then `QuantizerDepthwiseBase` and `QuantizerSeparableBase`. `BaseLayer` was the first step to give them all a common base class. I'll consolidate them (i.e. remove the unnecessary abstractions)  in a future PR.

jneeven:
Any idea what's up with codecov? How can project coverage go down if patch coverage is 100% :thinking:

lgeiger:
> Any idea what's up with codecov? How can project coverage go down if patch coverage is 100% 🤔

I removed a few lines which where previously covered, so the relative coverage can go down.
We might want to set a threshold for codecov before it complains.

PR #437, opened 2020-02-28
Simplify quantizer metrics

### LINKED ISSUE #410: Merge `QuantizerBase` and `BaseLayer`

lgeiger:
Closes #410

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -557,6 +557,14 @@ def get(identifier):
 
 
 def get_kernel_quantizer(identifier):
+    """Returns a quantizer from identifier and adds default kernel quantizer metrics.
+
+    # Arguments
+    identifier: Function or string
+
+    # Returns
+    `Quantizer` or `None`
+    """
     quantizer = get(identifier)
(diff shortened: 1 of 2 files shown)


larq/quantizers.py
```diff
@@ -554,3 +554,10 @@ def get(identifier):
     raise ValueError(
         f"Could not interpret quantization function identifier: {identifier}"
     )
+
+
+def get_kernel_quantizer(identifier):
```
leonoverweel: Do we want to make this a private method? (And add a docstring?)

lgeiger: I added a docstring.

I don't think we should make it a private method since we use it in `layers_base.py`  which is outside of this file.

PR #438, opened 2020-02-28
Add support for non-zero padding

lgeiger:
This adds support for non zero padding for `QuantConv1D`, `QuantConv2D`, `QuantConv3D`, `QuantDepthwiseConv2D`, `QuantSeparableConv1D `and `QuantSeparableConv2D` in order to better support https://github.com/larq/compute-engine/pull/252

We can add support for locally connected layers and transposed convolutions in the future.

### DIFF
--- a/larq/layers.py
+++ b/larq/layers.py
@@ -121,7 +121,7 @@ class QuantDense(QuantizerBase, tf.keras.layers.Dense):
 
 
 @utils.register_keras_custom_object
-class QuantConv1D(QuantizerBaseConv, QuantizerBase, tf.keras.layers.Conv1D):
+class QuantConv1D(QuantizerBase, QuantizerBaseConv, tf.keras.layers.Conv1D):
     """1D quantized convolution layer (e.g. temporal convolution).
 
     This layer creates a convolution kernel that is convolved with the layer input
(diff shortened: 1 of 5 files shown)


larq/layers.py
```diff
@@ -181,6 +182,7 @@ def __init__(
         kernel_size,
         strides=1,
         padding="valid",
+        pad_values=0.0,
```
lgeiger: @larq/core how do you feel about `pad_values` vs `padding_values`?

AdamHillier: I'm fine with either, no particular preference.

larq/layers_base.py
```diff
@@ -79,6 +88,79 @@ def get_config(self):
...
+    def build(self, input_shape):
+        if self._is_native_padding():
+            super().build(input_shape)
+        else:
+            with utils.patch_object(self, "padding", "valid"):
+                super().build(self._get_padding_same_shape(input_shape))
+
+    def call(self, inputs):
+        if self._is_native_padding():
+            return super(QuantizerBaseConv, self).call(inputs)
```
lgeiger: Unfortunately this Python 2 syntax is needed to make autograph in TF 1.14 happy.

lgeiger: 6b80b1d had the nice side-effect that we can get rid of this workaround 🎉

lgeiger:
> Only thing is I would prefer to test whether the layer actually uses non-zero padding (if I'm correct the test would now pass even if _is_native_padding() always returns True)

Good point. I added a check to make sure `tf.pad` gets called with the correct values  in f03c7142aa794afaf1a889cc937934a20f894da1

PR #439, opened 2020-03-02
Add method for printing human-readable memory sizes

leonoverweel:
This PR adds two things:

* Human-readable memory sizes are now no longer printed in megabytes: if they are smaller than 1 megabyte, they're printed in kilobytes; if they're larger than 1024 megabytes, they're printed in gigabytes.
* Changes abbreviations from e.g. `MB` to `MiB` to avoid any and all confusion about the relation between different orders of magnitude. (`MB` is sometimes interpreted as 1000 bytes instead of 1024 bytes.) [Relevant wiki](https://en.wikipedia.org/wiki/Binary_prefix); see table on the right.


larq/models.py
```diff
@@ -81,6 +81,27 @@ def _number_as_readable_str(num: float) -> str:
     return num + unit
 
 
+def _memory_as_readable_str(num_bits: int) -> str:
+    """Generate a human-readable string for the memory size.
+
+    1 KiB = 1024 B; we use the binary prefix (KiB) [1] instead of the decimal prefix
+    (KB) to avoid any confusion with multiplying by 1000 instead of 1024.
+
+    [1] https://en.wikipedia.org/wiki/Binary_prefix
```
jamescook106: Please could you replace this with a non-wikipedia reference?

larq/models.py
```diff
@@ -81,6 +81,27 @@ def _number_as_readable_str(num: float) -> str:
...
+    """
+
+    suffixes = ["B", "KiB", "MiB", "GiB", "TiB", "PiB", "EiB"]
```
timdebruin: By the time we're measuring the memory in exabytes we can be pretty sure something has gone wrong :rofl:

larq/models_test.py
```diff
@@ -128,6 +128,89 @@ def test_number_as_readable_str_large():
...
+        "128.00 EiB",
+        "256.00 EiB",
+        "512.00 EiB",
+        "1,024.00 EiB",
+        "2,048.00 EiB",
+    ]
+
+    for i, correct_string in enumerate(correct_strings):
+        assert lq.models._memory_as_readable_str(2 ** i) == correct_string
+
```
jamescook106: I like the idea of having another test that checks if a random amount of bits is covered successfully

larq/models.py
timdebruin: Would it make sense to only include the sensible suffixes such that if something *has* gone wrong the number at least looks big (without looking at the suffix)?

leonoverweel: I can get behind this. Should I go up to `GiB` then? Or even just `MiB`?

leonoverweel: What else do you think would break @timdebruin? (Besides Zoo snapshots, which we should be able to update fairly easily when Zoo gets bumped to whatever version of Larq this is released in.)

larq/models_test.py
leonoverweel: What do you mean exactly? A hardcoded "12,345 bits = x KiB" test or just checking whether passing any random int does not crash the method?

larq/models.py
leonoverweel: Done in e31028f.

larq/models.py
timdebruin: I am not sure, I added a test which built a network and suddenly @jamescook106 's tests broke because the model summaries had changed. But again I know very little about this which is why I tagged @jamescook106 . Maybe @AdamHillier or @lgeiger could clarify.

larq/models.py
```diff
@@ -81,6 +81,28 @@ def _number_as_readable_str(num: float) -> str:
...
+    num_bytes = num_bits / 8
+
+    for i, suffix in enumerate(suffixes):
+        rounded = num_bytes / (1024 ** i)
+        if rounded < 1024 or i == len(suffixes) - 1:
+            readable_str = f"{rounded:,.2f} {suffix}"
+            break
```
AdamHillier:
```suggestion
        if rounded < 1024:
            break
    readable_str = f"{rounded:,.2f} {suffix}"
```

larq/models.py
```diff
@@ -81,6 +81,27 @@ def _number_as_readable_str(num: float) -> str:
...
+
+    suffixes = ["B", "KiB", "MiB", "GiB", "TiB", "PiB", "EiB"]
+    num_bytes = num_bits / 8
+
+    for i, suffix in enumerate(suffixes):
+        rounded = num_bytes / (1024 ** i)
+        if rounded < 1024 or i == len(suffixes) - 1:
+            break
+
+    return f"{rounded:,.2f} {suffix}"
```
AdamHillier:
```suggestion
    for suffix in suffixes:
        if num_bytes < 1024:
            break
        num_bytes /= 1024

    return f"{num_bytes:,.2f} {suffix}"
```

leonoverweel: This breaks for the last suffix, so I think we should just stick with the current (working) version, even if it's not the prettiest.

larq/models_test.py
leonoverweel: This is also covered by the snapshot tests which have more "random" numbers.

PR #440, opened 2020-03-03
Round to 3 s.f. below GB, 0 d.p. above GB.

AdamHillier:
I think 3 s.f. is nicer than 2 d.p and is consistent with the `_num_as_readable_str` method which we use for op counts, et cetera.


leonoverweel:
Should we merge #439 first?

AdamHillier:
> Should we merge #439 first?

This is a PR into your PR branch, not into master.

AdamHillier:
Though that said I don't hugely mind making this a PR into master after your PR is merged.

leonoverweel:
Ah, gotcha. I'm fine with merging into my PR.

What's the motivation for rounding to three sig-figs instead of always two decimals? I think significant figures make more sense when there's a margin of error in measurements. However, we know this number exactly.

(I'm not opposed to this per se, and not terribly attached to the two-decimal rounding, but still interested to hear why this would be preferred.)

AdamHillier:
> What's the motivation for rounding to three sig-figs instead of always two decimals? I think significant figures make more sense when there's a margin of error in measurements. However, we know this number exactly.

Tbh I just think it looks nicer, and is 'purer' in some way. Also it is consistent with how we print numbers. But I also don't care that much if others don't have strong opinions.

I do think we should cap the size at GB though as mentioned in one of the PR comments.

leonoverweel:
Yes I have a commit ready on my branch to limit it to `GiB`; I think it makes more sense to keep this PR for the significant figures. What does everyone else think?

jneeven:
I have no opinion on this. Either are fine by me.

PR #445, opened 2020-03-04
Add Larq meta info to README.md

leonoverweel:
As discussed with @rnusselder, this PR adds some meta info to the Larq README:

* The purpose of the message at the top is to help people more easily discover (and click through to) the entire Larq ecosystem of libraries if they happen to stumble upon on one of the repos.
* The purpose of the About section is to point potential customers who are looking at Larq towards Plumerai, so that they can figure out who to contact from there. (We intentionally did not yet link to the Plumerai website.)

The plan is to add similar info to the Zoo and Compute Engine READMEs.


leonoverweel:
(I'll wait with the other PRs until this one is merged in case we tweak the copy.)

PR #446, opened 2020-03-04
Case opt check

leonoverweel:
Added in `_compute_var_opt_mapping()` so that the check only happens once.

### DIFF
--- a/larq/optimizers.py
+++ b/larq/optimizers.py
@@ -210,6 +210,13 @@ class CaseOptimizer(tf.keras.optimizers.Optimizer):
                         f"No `default_optimizer` provided to train variable `{var}`."
                     )
 
+        # Make sure that each optimizer touches at least one variable
+        for optimizer_index, (_, optimizer) in enumerate(self.pred_opt_pairs):
+            if not optimizer_index in self.var_opt_mapping.values():
+                raise ValueError(
+                    f"Optimizer `{optimizer}` did not claim any variables."
+                )
+
 
 @utils.register_keras_custom_object
 class Bop(tf.keras.optimizers.Optimizer):
--- a/larq/optimizers.py
+++ b/larq/optimizers.py
@@ -210,6 +210,9 @@ class CaseOptimizer(tf.keras.optimizers.Optimizer):
                         f"No `default_optimizer` provided to train variable `{var}`."
                     )
 
+        print("!~!!!!")
+        print(self.var_opt_mapping)
+
         # Make sure that each optimizer touches at least one variable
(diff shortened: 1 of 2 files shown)


larq/optimizers.py
```diff
@@ -210,6 +210,16 @@ def _compute_var_opt_mapping(self, grads_and_vars):
...
+        print("!~!!!!")
+        print(self.var_opt_mapping)
+
+        # Make sure that each optimizer touches at least one variable
+        for optimizer_index, (_, optimizer) in enumerate(self.pred_opt_pairs):
+            if optimizer_index not in self.var_opt_mapping.values():
+                raise ValueError(
+                    f"Optimizer `{optimizer}` did not claim any variables."
+                )
+
```
lgeiger: I think checking this in [`_apply_gradients()`](https://github.com/larq/larq/blob/c44ed976812edc2cc7d60c98155513942fc3af98/larq/optimizers.py#L136) would be simpler, then you could just do:
```python
for optimizer, opt_grads_and_vars in zip(self.optimizers, grad_var_lists):
    if not grad_var_lists:
        raise ValueError(f"Optimizer `{optimizer}` did not claim any variables.")
```

leonoverweel: Yes but then it happens on every training step; this way it just happens once when the var -> optimizer gets computed.

lgeiger: TensorFlow also raises errors in `_apply_gradients` too so it wouldn't be problematic, but this is fine too.

larq/optimizers_test.py
```diff
@@ -82,7 +82,40 @@ def test_overlapping_predicates(self):
     def test_missing_default(self):
         with pytest.warns(Warning):
```
lgeiger: Shouldn't this be?
```suggestion
        with pytest.raises(ValueError):
```

leonoverweel: Nope, we [warn about this](https://github.com/larq/larq/blob/master/larq/optimizers.py#L209) instead of throwing.

leonoverweel: I think would be for scenario for example where you "freeze" FP weights by only passing an optimizer for binary weights.

larq/optimizers.py
leonoverweel: I meant it mostly for efficiency.

larq/optimizers_test.py
lgeiger: :+1: I think I misread the comment [here](https://github.com/larq/larq/pull/446/files#diff-4930c450271ff44e3368d2a51994982eR110)

lgeiger: Sorry the git diff just was weird to read.

PR #447, opened 2020-03-05
Remove unnecessary # from format string

leonoverweel:
This will fix Atom and GitHub code highlighting.


AdamHillier:
> @koenhelwegen I think you originally added this, do you know if this is safe to remove?

I think I did, and I think it should be safe.

PR #449, opened 2020-03-06
Model size for 8-bit quantized full-precision weights

### LINKED ISSUE #448: Add int8 quantized size to model summary
### Feature motivation
Our model summary currently reports the following two metrics for model size:

```
| Model size:                        151.80 KiB |
| Float-32 Equivalent                158.91 KiB |
```

This is calculated as follows for models that contain both binary and full precision (FP) weights:

| Metric | Binary weights | FP weights |
| --- | --- | --- |
| `Model size` | counted as 1 bit | counted as 32 bits |
| `Float-32 Equivalent` | counted as 32 bits | counted as 32 bits |

For models in which FP weights will be 8-bit quantized for inference, it'd be useful to also know how large that model would be.

### Feature description
I propose adding another metric to the model summary:

| Metric | bin weights | FP weights |
| --- | --- | --- |
| `int-8 quantized equivalent` | counted as 1 bit | counted as 8 bits |

### Feature implementation
Adding another property to [`ModelProfile`](https://github.com/larq/larq/blob/master/larq/models.py#L242), similar to:

```python
    @property
    def fp_equivalent_memory(self) -> int:
        return sum(l.fp_equivalent_memory for l in self.layer_profiles)
```

And then adding it to the model summary printout.

leonoverweel:
This PR implements a new model summary property: model size if full-precision weights are quantized to be `int8` instead of `float32`; e.g. if they take up 1 byte of space instead of 4.

I'm also doing this for 16-bit weights in case the model is trained using half precision, but I'm not sure if this is necessary--the weights could still be stored as 32 bits by TensorFlow. If that's the case I'm happy to remove the second case in the if statement in the `int8_fp_weights_memory` property.

Closes #448.

### DIFF
--- a/larq/models.py
+++ b/larq/models.py
@@ -110,6 +110,14 @@ class WeightProfile:
     def fp_equivalent_memory(self) -> int:
         return 32 * self.count
 
+    @property
+    def int8_fp_weights_memory(self) -> int:
+        """Count any 32- or 18-bit weights as 8 bits instead."""
+
+        if self.bitwidth == 32 or self.bitwidth == 16:
+            return self.count * 8
+        return self.bitwidth * self.count
+
     def is_bias(self) -> bool:
         return "bias" in self._weight.name
 
@@ -147,6 +155,10 @@ class LayerProfile:
     @property
     def memory(self) -> int:
         return sum(p.memory for p in self.weight_profiles)
+
+    @property
+    def int8_fp_weights_memory(self) -> int:
+        return sum(p.int8_fp_weights_memory for p in self.weight_profiles)
 
     @property
     def fp_equivalent_memory(self) -> int:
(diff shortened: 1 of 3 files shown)


larq/models.py
```diff
@@ -110,6 +110,14 @@ def memory(self) -> int:
     def fp_equivalent_memory(self) -> int:
         return 32 * self.count
 
+    @property
+    def int8_fp_weights_memory(self) -> int:
+        """Count any 32- or 16-bit weights as 8 bits instead."""
+
+        if self.bitwidth == 32 or self.bitwidth == 16:
```
koenhelwegen: Would consider 
```suggestion
        if self.bitwidth > 8:
```
It doesn't really make sense to assume 32 bit weights will be quantized to 8 bits, while 24 bit weights remain 24.

PR #451, opened 2020-03-10
Add CI job to test larq against TensorFlow 2.2rc0

lgeiger:


lgeiger:
> Do you think these test failures are stuff we should debug ourselves now or wait for updates to the beta that may fix them first?

Our unittests relied on broken behaviour which was fixed in https://github.com/tensorflow/tensorflow/commit/81b21433716e4fad1d51c85219406ace2b6176c0#diff-7c464547ccc6362ecbb81b73b4cfbad0. This should be good to go now.

PR #454, opened 2020-03-11
Batchnorm in profile

koenhelwegen:
Currently, the profile simply counts BN weights, usually resulting in 4 full-precision variables per batch norm channel. However, for inference these are fused into 2. As these variables are full-precision, this can have a noticeable impact on total model size, especially in deep networks with lots of batch norms. This PR fixes this by only profiling relevant weights in the BN layer.

(This is a bit of a slippery slope, we can't anticipate all conversion-time optimizations in the profile. However, given that this particular optimization is both very common and easy to accommodate for, it's worth fixing.)

### DIFF
--- a/larq/models.py
+++ b/larq/models.py
@@ -2,12 +2,11 @@ import itertools
 from dataclasses import dataclass
 from typing import Any, Callable, Iterator, Mapping, Optional, Sequence, TypeVar, Union
 
+import larq.layers as lq_layers
 import numpy as np
 import tensorflow as tf
-from terminaltables import AsciiTable
-
-import larq.layers as lq_layers
 from larq.utils import memory_as_readable_str
+from terminaltables import AsciiTable
 
 __all__ = ["summary"]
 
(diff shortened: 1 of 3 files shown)


larq/models.py
```diff
@@ -2,12 +2,11 @@
...
 from typing import Any, Callable, Iterator, Mapping, Optional, Sequence, TypeVar, Union
 
+import larq.layers as lq_layers
 import numpy as np
 import tensorflow as tf
-from terminaltables import AsciiTable
-
-import larq.layers as lq_layers
 from larq.utils import memory_as_readable_str
+from terminaltables import AsciiTable
```
AdamHillier: This is an isort error.

larq/models_test.py
```diff
@@ -1,10 +1,10 @@
+import larq as lq
 import numpy as np
-import pytest
 import tensorflow as tf
-
-import larq as lq
 from larq.models import ModelProfile
 
+import pytest
+
```
AdamHillier: This is also an isort error.

larq/models.py
```diff
@@ -132,11 +131,20 @@ class OperationProfile:
 class LayerProfile:
     def __init__(self, layer: tf.keras.layers.Layer):
         self._layer = layer
+
+        weights = layer.weights
+        if isinstance(layer, tf.keras.layers.BatchNormalization):
+            fused_pairs = [("beta", "moving_mean"), ("gamma", "moving_variance")]
+            for pair in fused_pairs:
+                names = [w.name.split("/")[-1].replace(":0", "") for w in weights]
```
AdamHillier: This is quite brittle because it relies on how BN in Keras names its variables, so this line could throw, but as it's in the test I think it's okay.

PR #461, opened 2020-03-20
Support unknown input shapes with one padding

lgeiger:
This makes sure that `Layer.build` doesn't throw when input shapes are undefined.

### DIFF
--- a/larq/layers_base.py
+++ b/larq/layers_base.py
@@ -14,7 +14,7 @@ def _compute_padding(stride, dilation_rate, input_size, filter_size):
     effective_filter_size = (filter_size - 1) * dilation_rate + 1
     output_size = (input_size + stride - 1) // stride
     total_padding = (output_size - 1) * stride + effective_filter_size - input_size
-    total_padding = total_padding if total_padding > 0 else 0
+    total_padding = max(total_padding, 0)
     padding = total_padding // 2
     return padding, padding + (total_padding % 2)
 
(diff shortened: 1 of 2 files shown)


larq/layers_base.py
```diff
@@ -14,7 +14,7 @@ def _compute_padding(stride, dilation_rate, input_size, filter_size):
     effective_filter_size = (filter_size - 1) * dilation_rate + 1
     output_size = (input_size + stride - 1) // stride
```
koenhelwegen: This line still throws ``TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'``

lgeiger: OK, looks like we can also have dynamic shapes in `Layer.call`. I changed it to pass a Tensor.

larq/layers_base.py
```diff
@@ -101,17 +101,11 @@ def _is_native_padding(self):
...
-            for stride, dilation_rate, input_size, filter_size in zip(
-                self.strides, self.dilation_rate, shape, self.kernel_size
-            )
-        )
-
-    def _get_spatial_padding_same_shape(self, shape):
         return [
-            size + sum(pad)
-            for size, pad in zip(shape, self._get_spatial_padding_same(shape))
+            _compute_padding(stride, dilation_rate, shape[i], filter_size)
```
lgeiger: Unfortunately Tensors don't support iteration in Graph model, so we need to do `shape[i]` instead.

PR #463, opened 2020-03-24
Use static tensor shapes if possible for one-padding

lgeiger:
In general the batch dimension is undefined. This means that [`tf.pad`](https://github.com/tensorflow/tensorflow/blob/r2.1/tensorflow/python/ops/array_ops.py#L2952-L2968) doesn't correctly set the shapes of the tensor. This makes model summary fail since the `TensorShape` is not correctly inferred.

This PR uses static shapes whenever the spatial dimensions are fully defined to avoid this problem.

### DIFF
--- a/larq/layers_base.py
+++ b/larq/layers_base.py
@@ -116,7 +116,9 @@ class QuantizerBaseConv(tf.keras.layers.Layer):
         )
 
     def _get_padding_same(self, inputs):
-        input_shape = tf.shape(inputs)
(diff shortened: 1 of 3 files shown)


lgeiger:
Not sure why codecov doesn't like this, the test should fully cover the changes.

AdamHillier:
Perhaps we should add an example with 1-padding to the model summary snapshot tests to detect this in the future?

PR #466, opened 2020-03-28
Add support for experimental_aggregate_gradients

lgeiger:
This PR adds support for `experimental_aggregate_gradients` in `apply_gradients` for `Bop` and the `CaseOptimizer` which will be introduced in TensorFlow 2.2

[`experimental_aggregate_gradients=False`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L434-L473) allows users to aggregate gradients themselfs. E.g. this is use by [Keras to manually aggregate scaled gradients in fp16 when using multi-GPU mixed precision training before passing them to the optimizer](https://github.com/tensorflow/tensorflow/blob/b97c90c87fc433140bae47d93521a3b795645c12/tensorflow/python/keras/engine/training.py#L1822-L1845).

Adding support for it also works around a bug in TF 2.2-rc2 where in mixed precision training [`_HAS_AGGREGATE_GRAD` is not checked so `experimental_aggregate_gradients` is always passed to the wrapped optimizer](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/mixed_precision/experimental/loss_scale_optimizer.py#L275-L282) which would fail with a cryptic error message otherwise.

Since the keyword argument might change in the future, I just forward `**kwargs` to the underlying optimizers.

This PR also adds the correct name scopes in `apply_gradients` which previously weren't used.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tf-version: [1.14.0, 1.15.2, 2.0.1, 2.1.0, 2.2.0rc0]
+        tf-version: [1.14.0, 1.15.2, 2.0.1, 2.1.0, 2.2.0rc2]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


lgeiger:
> So this makes us ready for half-precision training in TF2.2?

Yes, this works around some issues in the rc.
For version < 2.2, this PR includes no functional changes.

PR #468, opened 2020-03-30
Make quantiser names consistent. Closes #467.

### LINKED ISSUE #467: Inconsistent quantiser names
This is very minor and I'm very happy for this to be a `wontfix`.

At the moment two out of eight quantizers have a "Quantizer" suffix:

image
It seems to me it should be all of them with the suffix or none of them with the suffix. We could potentially standardise on one version but have an alias to the other version.

AdamHillier:


larq/quantizers.py
```diff
@@ -52,14 +52,16 @@
...
-    "SteHeaviside",
+    "DoReFa",
     "DoReFaQuantizer",
+    "MagnitudeAwareSign",
+    "NoOp",
     "NoOpQuantizer",
+    "SteHeaviside",
+    "SteSign",
+    "SteTern",
+    "SwishSign",
```
AdamHillier: These have just been re-ordered alphabetically.

AdamHillier:
> I think we mention `NoOpQuantizer` a few times in the docs. Could you change that to recommend `NoOp` instead?

Thanks for spotting that, should be done in #4526cb6.

larq/quantizers.py
```diff
@@ -536,7 +536,7 @@ class DoReFa(BaseQuantizer):
     \end{cases}\\]
 
     ```plot-activation
-    quantizers.DoReFaQuantizer
+    quantizers.DoReFa
```
AdamHillier: Will this break something?

lgeiger: That shouldn't break things :+1:

PR #482, opened 2020-05-04
Reformat docs for keras-autodoc

leonoverweel:
This adds indentations to docstrings so that [keras-autodoc](https://github.com/keras-team/keras-autodoc) understands sections with lists of arguments, input shapes, output shapes, returns, etc. See larq/docs#68.

Before:

```
# Arguments:
arg_name: this is the description.
```

After:

```
# Arguments:
    arg_name: this is the description.
```

We'll probably need to do a minor release of Larq once we do this, and then bump the version that larq/docs uses to that version.

### DIFF
--- a/larq/layers.py
+++ b/larq/layers.py
@@ -63,28 +63,29 @@ class QuantDense(QuantizerBase, tf.keras.layers.Dense):
         ```
 
     # Arguments
-    units: Positive integer, dimensionality of the output space.
-    activation: Activation function to use. If you don't specify anything,
-        no activation is applied (`a(x) = x`).
-    use_bias: Boolean, whether the layer uses a bias vector.
-    input_quantizer: Quantization function applied to the input of the layer.
-    kernel_quantizer: Quantization function applied to the `kernel` weights matrix.
-    kernel_initializer: Initializer for the `kernel` weights matrix.
-    bias_initializer: Initializer for the bias vector.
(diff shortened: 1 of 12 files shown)


larq/layers.py
```diff
@@ -820,58 +837,59 @@ class QuantConv2DTranspose(QuantizerBase, tf.keras.layers.Conv2DTranspose):
...
-    `(batch, new_rows, new_cols, filters)` if data_format='channels_last'.
-    `rows` and `cols` values might have changed due to padding.
+        4D tensor with shape:
+        `(batch, filters, new_rows, new_cols)` if data_format='channels_first'
+        or 4D tensor with shape:
+        `(batch, new_rows, new_cols, filters)` if data_format='channels_last'.
+        `rows` and `cols` values might have changed due to padding.
 
     # References
     - [A guide to convolution arithmetic for deep
```
lgeiger, line 895: It looks like this needs another indentaion

leonoverweel: Oops, fixed last night but hadn't pushed yet.

leonoverweel:
I haven't been able to find one, though I agree that it'd be very nice to have.

Let's chat on Monday about coordinating merging this with a docs release?

lgeiger:
> Let's chat on Monday about coordinating merging this with a docs release?

Sounds good. Since we have a few unreleased bugfixes it would be good to get a release out anyway.

PR #491, opened 2020-05-18
Correctly compute one-padding shape to prevent recreating conv_op

lgeiger:
#461 introduced a bug which broke the computation of the padded output shape with one-padding. This resulted in convolution ops being recreated after the layer has been built. This PR fixes it by correctly computing the padded shape so that convolutional ops are not recreated after the layer has been built.

I noticed this bug when trying to debug @jneeven multi-GPU issues with Quantizer that call `self.add_update` and `one-padding`.

### DIFF
--- a/larq/layers_base.py
+++ b/larq/layers_base.py
@@ -10,11 +10,20 @@ from larq.quantizers import Quantizer
 log = logging.getLogger(__name__)
 
 
-def _compute_padding(stride, dilation_rate, input_size, filter_size):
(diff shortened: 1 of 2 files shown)


lgeiger:
Looks like codecov get's confused with TensorFlow autograph, I checked and the code that it is complaining about is actually covered by the existing unittests.

jneeven:
> Looks like codecov get's confused with TensorFlow autograph, I checked and the code that it is complaining about is actually covered by the existing unittests.

Overall coverage seems to not have changed, so I think we can just merge this?

Issue #487, opened 2020-05-09
Training does not converge when DoReFa is used as a kernel quantizer


bferrarini:
### Describe the bug
Hi,
I tried to use DoReFa quantiser to train a simple model for CIFAR10, but training failed to converge:
```
10000/1 - 1s 96us/sample - loss: 2.3026 - accuracy: 0.1000
Test loss: 2.3025851249694824
Test accuracy: 0.1
```
I found that the problem only occurs when DoReFa quantizer is used on the kernel (kernel_quantizer)

### This code reproduces the problem. DoReFa is used for both activations and Weights.
```
from __future__ import print_function
import tensorflow as tf
from tensorflow.keras.datasets import cifar10
from tensorflow.keras.models import Sequential
from tensorflow.keras.layers import Activation, Flatten, BatchNormalization
from tensorflow.keras.layers import MaxPooling2D
import larq as lq
from larq.layers import QuantConv2D, QuantDense
import time

import os
from larq.quantizers import DoReFaQuantizer

batch_size = 128
num_classes = 10
epochs = 30
num_predictions = 20
save_dir = os.path.join(os.getcwd(), 'saved_models')

clip_value = 1.0
model_name = 'keras_cifar10_trained_model_test_DOREFA'

TEST = False

(x_train, y_train), (x_test, y_test) = cifar10.load_data()
print('x_train shape:', x_train.shape)
print(x_train.shape[0], 'train samples')
print(x_test.shape[0], 'test samples')

y_train = tf.keras.utils.to_categorical(y_train, num_classes)
y_test = tf.keras.utils.to_categorical(y_test, num_classes)


model = Sequential()
model.add(QuantConv2D(32, (3, 3), padding='same',
                      input_quantizer = None,
                      kernel_quantizer = DoReFaQuantizer(),
                      kernel_constraint = lq.constraints.WeightClip(clip_value=1),
                      use_bias = True,
                 input_shape=x_train.shape[1:]))
model.add(BatchNormalization(momentum=0.9))

model.add(QuantConv2D(32, (3, 3), padding='valid',
                      input_quantizer = DoReFaQuantizer(),
                      kernel_constraint=lq.constraints.WeightClip(clip_value=1),
                      use_bias = False,
                      kernel_quantizer = DoReFaQuantizer()))      
model.add(MaxPooling2D(pool_size=(2, 2)))
model.add(BatchNormalization(momentum=0.9))

model.add(QuantConv2D(64, (3, 3), padding='same',
[...]

PR #494, opened 2020-05-19
Add warning to DoReFa docstring

jneeven:
to make clear that it is only valid for activations. See #487 .


larq/quantizers.py
```diff
@@ -48,7 +48,6 @@
 from typing import Callable, Union
 
 import tensorflow as tf
```
jneeven, line 50:
```suggestion
import tensorflow as tf

```

PR #497, opened 2020-05-29
Fix metric typo

koenhelwegen:
This corrects a very minor typo that didn't show up in the tests because we were explicitly building. However, this did make the [docstring example](https://github.com/larq/larq/blob/04ea8153ef10484dab68873cfe3ea37db5615ca4/larq/metrics.py#L21) fail.

### DIFF
--- a/larq/metrics.py
+++ b/larq/metrics.py
@@ -36,12 +36,6 @@ class FlipRatio(tf.keras.metrics.Metric):
         self.built = False
         self.values_dtype = tf.as_dtype(values_dtype)
 
-    def __call__(self, inputs, **kwargs):
-        if not self.built:
-            with tf.name_scope(self.name), tf.init_scope():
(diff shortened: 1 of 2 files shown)


larq/metrics.py
```diff
@@ -36,12 +36,6 @@ def __init__(self, values_dtype="int8", name="flip_ratio", dtype=None):
         self.built = False
         self.values_dtype = tf.as_dtype(values_dtype)
 
-    def __call__(self, inputs, **kwargs):
-        if not self.built:
-            with tf.name_scope(self.name), tf.init_scope():
-                self.build(inputs.shape)
-        return super().__call__(inputs, **kwargs)
```
lgeiger, line 43: For reference `super().__call__` will implicitely call `update_state` so we don't need to overwrite it here.

PR #498, opened 2020-05-30
Add support for TensorFlow 2.3

lgeiger:
This integrates changes of https://github.com/tensorflow/tensorflow/commit/8d31fb4b765260a6d0bcaf825ca68766b5b798ee#diff-8db8dd75efb914430be51c632dfa452e and https://github.com/tensorflow/tensorflow/commit/c5caa29b5e6d10079020673a0dbd0035214df94d

~~Since these changes haven't been released yet I didn't add any CI checks that test against tf-nightly~~ I added TF 2.3.0rc0 to the CI test matrix.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,7 +18,7 @@ jobs:
           python-version: 3.7
       - name: Install project dependencies
         run: |
-          pip install tensorflow==2.1.0
+          pip install tensorflow-cpu==2.2.0
           pip install -e .[test]
       - name: Run Flake8
         run: flake8
--- a/larq/quantized_variable.py
+++ b/larq/quantized_variable.py
@@ -3,16 +3,16 @@
 from typing import Optional
 
 import tensorflow as tf
-from tensorflow.python.distribute.values import DistributedVariable
+from tensorflow.python.distribute.values import DistributedVariable  # type: ignore
 from tensorflow.python.framework import ops
 from tensorflow.python.ops import resource_variable_ops
 
 from larq import context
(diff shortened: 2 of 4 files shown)


.github/workflows/lint.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           python-version: 3.7
       - name: Install project dependencies
         run: |
-          pip install tensorflow==2.1.0
+          pip install tensorflow-cpu==2.2.0
```
leonoverweel, line 21: Are we always linting on `tensorflow==2.latest-1.0`?

larq/quantized_variable.py
```diff
@@ -360,4 +364,7 @@ def _as_graph_element(self):
 tf.register_tensor_conversion_function(
     QuantizedVariable, QuantizedVariable._dense_var_to_tensor
 )
-ops.register_dense_tensor_like_type(QuantizedVariable)
+try:
+    ops.register_dense_tensor_like_type(QuantizedVariable)
```
leonoverweel, line 368: So this is for `tensorflow<=2.2.0` and no longer needed because of the new `core.Tensor` baseclass in `>=2.3.0rc0`?

.github/workflows/lint.yml
lgeiger: I changed it to 2.2 in order to remove some pytype errors that where caused by the way 2.1 handled some imports.

larq/quantized_variable.py
lgeiger: Correct

PR #506, opened 2020-06-15
Cleanup tests

lgeiger:
This PR simplifies some of our testing code and switches to using `@pytest.mark.usefixtures` in cases where function arguments are never used. I think this makes it easier to read and understand what these arguments do. I also moved the data format parameterization into a shared fixture.

### DIFF
--- a/larq/conftest.py
+++ b/larq/conftest.py
@@ -73,3 +73,8 @@ def quantized(request):
     """pytest fixture for running test quantized and non-quantized"""
     with lq_context.quantized_scope(request.param):
         yield request.param
+
+
+@pytest.fixture(params=["channels_last", "channels_first"])
+def data_format(request):
+    return request.param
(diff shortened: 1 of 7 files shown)


larq/conftest.py
```diff
@@ -73,3 +73,8 @@ def quantized(request):
     """pytest fixture for running test quantized and non-quantized"""
     with lq_context.quantized_scope(request.param):
         yield request.param
+
+
+@pytest.fixture(params=["channels_last", "channels_first"])
+def data_format(request):
+    return request.param
```
jneeven, line 80: Could you explain what this does? From [the pytest docs](https://docs.pytest.org/en/stable/fixture.html#fixtures-can-introspect-the-requesting-test-context) I understand that `request.param` refers to some property of the calling function?

lgeiger: When using `data_format` as an argument to a testing funciton pytest will run the function twice, once with `request.param=="channels_last"` and once with `request.param=="channels_first"`. See [here](https://github.com/larq/larq/pull/506/files#diff-fd74c9d55bedda7f687491b4acd9e468L223-R223) for a usage example.

PR #485, opened 2020-05-06
Fix case optimizer when used with distributed mixed precision loss scaling

lgeiger:
This fixes some issues that can occur when using the `CaseOptimizer` in distributed training with the mixed precision `LossScaleOptimizer`.

The failure on master can be reproduced using:
```python
import tensorflow as tf
from tensorflow.python.keras import testing_utils
from tensorflow.python.distribute.strategy_combinations import set_virtual_cpus_to_at_least

import larq as lq
from larq import testing_utils as lq_testing_utils

set_virtual_cpus_to_at_least(3)  # Simulate distributed training on CPU only machine

(x_train, y_train), _ = testing_utils.get_test_data(
    train_samples=1000, test_samples=0, input_shape=(10,), num_classes=2
)
y_train = tf.keras.utils.to_categorical(y_train)


with tf.distribute.MirroredStrategy(["/cpu:1", "/cpu:2"]).scope():
    model = lq_testing_utils.get_small_bnn_model(
        x_train.shape[1], 20, y_train.shape[1], trainable_bn=True
    )

    opt = lq.optimizers.CaseOptimizer(
        (lq.optimizers.Bop.is_binary_variable, lq.optimizers.Bop()),
        default_optimizer=tf.keras.optimizers.Adam(0.01),
    )
    opt = tf.keras.mixed_precision.experimental.LossScaleOptimizer(opt, "dynamic")
    model.compile(loss="categorical_crossentropy", optimizer=opt, metrics=["acc"])

    model.fit(x_train, y_train, epochs=2, batch_size=16, verbose=0)
```

Adding a unittest for this is tricky since calling `set_virtual_cpus_to_at_least` needs proper cleanup of the TensorFlow test session.

Closes #396

PR #507, opened 2020-06-20
Run distribution strategy tests with multiple virtual CPUs

lgeiger:
This PR changes the distribution strategy fixture to use two virtual CPUs. This will correctly simulate distributed training which allows us to catch bugs like #485 in our unittests and allows for easy local debugging of multi-GPU issues.

I had to ignore a few known failure cases due to some underlying TensorFlow issues of quantized variables, since the new fixture correctly simulates multi-GPU environments on CI.

### DIFF
--- a/larq/conftest.py
+++ b/larq/conftest.py
@@ -1,9 +1,16 @@
 import pytest
 import tensorflow as tf
 from packaging import version
+from tensorflow.python.distribute import strategy_combinations
 from tensorflow.python.eager import context
 
 from larq import context as lq_context
+
+if version.parse(tf.__version__) >= version.parse("1.15"):
+    strategy_combinations.set_virtual_cpus_to_at_least(3)
+    distributed_devices = ["/cpu:1", "/cpu:2"]
+else:
+    distributed_devices = ["/cpu:0"]
 
 
 @pytest.fixture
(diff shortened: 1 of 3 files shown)


larq/conftest.py
```diff
@@ -1,10 +1,17 @@
 import pytest
 import tensorflow as tf
 from packaging import version
+from tensorflow.python.distribute import strategy_combinations
 from tensorflow.python.eager import context
 
 from larq import context as lq_context
 
+if version.parse(tf.__version__) >= version.parse("1.15"):
+    strategy_combinations.set_virtual_cpus_to_at_least(3)
```
lgeiger, line 10: This sets the device config globally for all unittests since otherwise we would need to properly reset the execution context after each unittest.

larq/quantized_variable_test.py
```diff
@@ -85,9 +85,10 @@ def test_method_delegations(distribute_scope):
             # These attributes are not supported for DistributedVariables
             assert x.constraint is None
             assert x.initializer == x.latent_variable.initializer
-        assert evaluate(x.assign(4)) == 8
-        assert evaluate(x.assign_add(1)) == 10
-        assert evaluate(x.assign_sub(1.5)) == 7
+            # DistributedVariables.assign doesn't correctly return quantized variables
+            assert evaluate(x.assign(4)) == 8
+            assert evaluate(x.assign_add(1)) == 10
+            assert evaluate(x.assign_sub(1.5)) == 7
```
koenhelwegen, line 91: Wouldn't it be nicer to split the assign and read instead of skipping?
```suggestion
        
        def apply_and_read(x, fn, args):
            evaluate(fn(*args))
            return evaluate(x)

        assert apply_and_read(x, x.assign, [4]) == 8
        assert apply_and_read(x, x.assign_add, [1]) == 10
        assert apply_and_read(x, x.assign_sub, [1.5]) == 7
```

PR #514, opened 2020-07-03
Fix saving in SavedModel format

lgeiger:
This PR fixes saving of larq models in the TensorFlow `SavedModel` format.

### DIFF
--- a/larq/quantized_variable_test.py
+++ b/larq/quantized_variable_test.py
@@ -305,10 +305,6 @@ def test_optimizer(should_quantize):
 @pytest.mark.skipif(
     version.parse(tf.__version__) < version.parse("2"), reason="Requires TensorFlow 2",
 )
-@pytest.mark.skipif(
-    version.parse(tf.__version__) == version.parse("2.3.0rc0"),
-    reason="Currently broken on 2.3.0rc0.",
-)
 def test_saved_model(tmp_path):
     model_path = str(tmp_path / "model")
     x = np.random.normal(size=(4, 32))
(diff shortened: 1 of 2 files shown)


larq/quantized_variable_test.py
```diff
@@ -299,3 +300,15 @@ def f():
             assert evaluate(x) == -2.0
     else:
         assert evaluate(x) == 0.0
+
+
+@pytest.mark.skipif(
+    version.parse(tf.__version__) < version.parse("2"), reason="Requires TensorFlow 2",
+)
```
lgeiger, line 307: I guess we could make this work in TF 1.x as well, but ignore the failure for now.

larq/quantized_variable_test.py
```diff
@@ -299,3 +300,16 @@ def f():
...
+@pytest.mark.skipif(version.parse(tf.__version__) == version.parse("2.3.0rc0"))
```
lgeiger, line 308: We should fix this in a future PR or open an issue on TensorFlow

larq/quantized_variable_test.py
```diff
@@ -299,3 +300,19 @@ def f():
...
     else:
         assert evaluate(x) == 0.0
+
+
+@pytest.mark.skipif(
+    version.parse(tf.__version__) < version.parse("2"), reason="Requires TensorFlow 2",
+)
+@pytest.mark.skipif(
+    version.parse(tf.__version__) == version.parse("2.3.0rc0"),
+    reason="Currently broken on 2.3.0rc0.",
```
leonoverweel, line 310: Should we make an issue tracking the TF 2.3 release to keep track of potential blockers/things to fix like this?

lgeiger: :+1: I will look into this soon and will either make an issue or provide a fix.

lgeiger: 2c30f2170cc84e8cd839d59018dbf865a0745f62 should fix the incompatibility with TF 2.3

lgeiger:
The codecov failures are expected since some codepaths only get called on specific versions of TensorFlow.

PR #515, opened 2020-07-03
Always upload coverage on CI to reduce flaky code coverage reports

lgeiger:
Since we have special handling for different TensorFlow version Codecov sometimes reported wrong results (see e.g.  #514). This PR changes our CI to always compute coverage and let Codecov handle the merging automatically.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -11,9 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tf-version: [1.14.0, 1.15.2, 2.0.1, 2.1.0, 2.3.0rc0]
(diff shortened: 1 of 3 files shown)


.github/workflows/unittest.yml
```diff
@@ -25,23 +27,9 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          pip install tensorflow==${{matrix.tf-version}}
+          pip install tensorflow-cpu==${{matrix.tf-version}} || pip install tensorflow==${{matrix.tf-version}}
```
lgeiger, line 30: This change is just to speedup CI if a CPU only version of TensorFlow is available.

PR #516, opened 2020-07-04
Upgrade isort

lgeiger:
See https://timothycrosley.github.io/isort/docs/major_releases/introducing_isort_5/

### DIFF
--- a/larq/quantized_variable.py
+++ b/larq/quantized_variable.py
@@ -9,12 +9,14 @@ from tensorflow.python.ops import resource_variable_ops
 from larq import context
 from larq.quantizers import Quantizer
 
+# pytype: disable=import-error
 try:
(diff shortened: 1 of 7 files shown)


larq/quantized_variable.py
```diff
@@ -11,8 +10,9 @@
 from larq.quantizers import Quantizer
 
 try:  # pragma: no cover
-    from tensorflow.python.types.core import Tensor as TensorType  # type: ignore
-    from tensorflow.python.distribute.ps_values import AggregatingVariable  # type: ignore
+    # type: ignore
```
AdamHillier, line 13: I thought these `# type: ignore` comments applied to the current line? Does this apply to the following line?

lgeiger: Good catch, this would actually disable silence type errors for the entire file. Fixed it in the last commit.

PR #556, opened 2020-08-17
Fix distributed quantized variable assignment in TF 2.x

lgeiger:
This PR adds support for multiple quantized variable assignments inside a distribution strategy. This edge case was previously not supported due to limitations in TensorFlow. This PR adapts the workaround of https://github.com/tensorflow/tensorflow/pull/41214 which makes `QuantizedVariables` behave consistently within a distribution strategy in eager execution. After this PR has been merged I'd be comfortable to expose `QuantizedVariables` as a public API to implement custom layers or training strategies.

### DIFF
--- a/larq/quantized_variable.py
+++ b/larq/quantized_variable.py
@@ -19,6 +19,9 @@ except ModuleNotFoundError:
 # pytype: enable=import-error
 
 
+UNSPECIFIED = object()
+
+
 class QuantizedVariable(tf.Variable, TensorType):
     """A Variable that can be quantized in the forward pass in applicable contexts."""
 
@@ -27,6 +30,7 @@ class QuantizedVariable(tf.Variable, TensorType):
         variable: tf.Variable,
         quantizer: Optional[Quantizer] = None,
         precision: Optional[int] = None,
+        op: Optional[tf.Operation] = UNSPECIFIED,
     ):
         """Creates an QuantizedVariable instance.
 
@@ -36,6 +40,7 @@ class QuantizedVariable(tf.Variable, TensorType):
                 variable to a fake quantized variable.
             precision: An optional integer defining the precision of the quantized
                 variable. If `None`, `quantizer.precision` is used.
+            op: An optional operation of this variable.
         """
         if not resource_variable_ops.is_resource_variable(variable):
             raise ValueError(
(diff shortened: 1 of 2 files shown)


larq/quantized_variable.py
```diff
@@ -27,6 +30,7 @@ def __init__(
         variable: tf.Variable,
         quantizer: Optional[Quantizer] = None,
         precision: Optional[int] = None,
+        op: Optional[tf.Operation] = UNSPECIFIED,
```
AdamHillier, line 33: I'm surprised this passes type checking, why does `UNSPECIFIED` (which is equal to `object()`) satisfy `Optional[tf.Operation]`?

Is there a problem using `None` as the default value here, or can `None` be interpreted as a meaningful `tf.Operation`?

AdamHillier: Ah I see what the issue is, you have to know whether or not to forward the `op` of the underlying variable or not, which is why you use a sentinel object rather than just `None`?

In that case maybe we could do something [like this](https://github.com/larq/zookeeper/blob/a7b1ddb8e054d9b0931278a176c0e872792841c7/zookeeper/core/utils.py#L10-L16) and define a class of the sentinel object, so that it can be referred to in the type hint rather than using `Optional`?

lgeiger: > Ah I see what the issue is, you have to know whether or not to forward the op of the underlying variable or not, which is why you use a sentinel object rather than just None?

Correct, I need to distinguish between `None` and unset. 

> In that case maybe we could do something like this and define a class of the sentinel object, so that it can be referred to in the type hint rather than using Optional?

I'd still need to use `Optional` or `Union[tf.Operation, None]` in both cases since `op` is explicitely allowed to be `None`. I mainly opted for this solution since it is widely used in the TensorFlow codebase.

> I'm surprised this passes type checking, why does `UNSPECIFIED` (which is equal to object()) satisfy `Optional[tf.Operation]`?

Yeah, not sure if this is a bug or expected behaviour. Maybe it will implicitely make a union of the specified type and the default value?

I have the feeling that `pytype` is simply broken, it doesn't even throw an error for me if I add the following to this file:
```python
def foo(i: int):
    return i

foo("3.0") # this is clearly wrong
```

jneeven (approved):
I can't say that I really understand everything that's going on here, but this PR is becoming somewhat stale so I'd be fine with merging.

AdamHillier (approved):

PR #577, opened 2020-09-25
Only warn for missing kernel constraints when quantizers are binary

lgeiger:
Larq can be used with many different quantizer and not all of them require a kernel constraint for proper training. This PR changes the warning message to only trigger when using a binary quantizer without a kernel constraint.


jneeven (approved):
Thank you 🙌  this has been annoying me for a while

PR #579, opened 2020-09-25
:arrow_up: 0.10.1

lgeiger:
We've had a few small fixes which I think would be nice to get released.


leonoverweel (approved):
Nice, sounds good.

PR #596, opened 2020-10-21
Add __version__ property. Closes #595.

### LINKED ISSUE #595: __version__
### Feature motivation
Is there a way to dynamically poll the version of larq (or lce or larq-zoo for that matter)?
If not, could it be done using `__version__` as usual for standard library modules?

### Feature description
```
import larq
print(larq.__version__)
```

AdamHillier:


AdamHillier:
Not sure how to test this, or rather it seems a bit clunky to create a new test file just for this one thing.

lgeiger (approved):
Looks good!

> Not sure how to test this, or rather it seems a bit clunky to create a new test file just for this one thing.

I'd be fine not testing it explicitely since it should be executed on each import anyway.

AdamHillier:
Cool, I added a small test file that checks for the presence of "." but I'm also happy to remove it.

PR #614, opened 2021-01-27
Support 1d ops in model profile

jneeven:
The model summary currently breaks for models with 1D convs, but the computation of the number of output pixels is exactly the same as in 2D (except only over two dimensions).

### DIFF
--- a/larq/models.py
+++ b/larq/models.py
@@ -11,21 +11,6 @@ from larq.utils import memory_as_readable_str
 
 __all__ = ["summary"]
 
-op_count_supported_layer_types = (
-    lq_layers.QuantConv2D,
-    lq_layers.QuantSeparableConv2D,
-    lq_layers.QuantDepthwiseConv2D,
-    lq_layers.QuantDense,
-    tf.keras.layers.Conv2D,
-    tf.keras.layers.SeparableConv2D,
-    tf.keras.layers.DepthwiseConv2D,
-    tf.keras.layers.Dense,
-    tf.keras.layers.Flatten,
-    tf.keras.layers.BatchNormalization,
-    tf.keras.layers.MaxPool2D,
-    tf.keras.layers.AveragePooling2D,
-)
-
 mac_containing_layers = (
     lq_layers.QuantConv2D,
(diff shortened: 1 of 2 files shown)


larq/models.py
```diff
@@ -239,8 +239,8 @@ def output_pixels(self) -> Optional[int]:
         """Number of pixels for a single feature map (1 for fully connected layers)."""
         if not self.output_shape:
             return None
-        if len(self.output_shape) == 4:
-            return int(np.prod(self.output_shape[1:3]))
+        if len(self.output_shape) == 3 or len(self.output_shape) == 4:
+            return int(np.prod(self.output_shape[1:]))
```
lgeiger, line 243: In the case of `len(self.output_shape) == 4`, this doesn't giove the same result since:
```python
l = [0, 1, 2, 3]
assert len(l) == 4
assert l[1:3] == [1, 2]
assert l[1:] == [1, 2, 3]
```
Or am I missing something?

lgeiger (changes requested):

larq/models.py
jneeven: I see, I locally had `[1:2]` but thought I could combine the two here and make it somewhat cleaner, but indeed this is incorrect :+1:

lgeiger: I see, then I guess the following should work:
```suggestion
            return int(np.prod(self.output_shape[1:-1]))
```

jneeven: It should, but I already fixed it by just adding another if-branch returning `shape[1]` for the 1D case, which is probably a bit cleaner than taking the `np.prod` of a one-length array

jneeven:
I'll add a 1D unit test for this soon

jneeven:
@lgeiger I have not explicitly tested the `LocallyConnected1D` layer. I tried adding a test for it and ensuring it returns `None` at various points, but that doesn't fit well with the current unit tests and it's an unlikely op to be used anyway. If we ever run into issues there, let's worry about it at that time.

lgeiger (approved):
Thanks for adding the unittests 🚀

Issue #620, opened 2021-02-12
build() call not forwarded via base class


Joschua-Conrad:
My use case: I develop a custom quantizer class, which derives from *larq.quantizers.BaseQuantizer* and which is then used like any other pre-defined quantizer. The custom quantizer defines a *build()* method.

Inside [BaseQuantizer.build](https://github.com/larq/larq/blob/094eaf1cca5017cdc643bf68bbb1301ab3f1af0f/larq/quantizers.py#L157) the method call to *build* is not forwarded via its base class *tf.keras.layers.Layer*.

In the [Tutorial about custom layers](https://www.tensorflow.org/guide/keras/custom_layers_and_models#best_practice_deferring_weight_creation_until_the_shape_of_the_inputs_is_known) this is also skipped. But when looking at [the build-method code in the base class](https://github.com/tensorflow/tensorflow/blob/85c8b2a817f95a3e979ecd1ed95bff1dc1335cff/tensorflow/python/keras/engine/base_layer.py#L448-L467), it is seen that the call should be forwarded, in order to set *self.built*, such that it is ensured, that the build triggered via *call* is only done once.

This is not an issue in most cases, as most users do not use explicit calls to *build* and use implicit builds triggered via *tf.keras.layers.Layer.call*, only. In my case, the quantizer is built explicitly before using *call* the first time and *call* tries to build the layer again, which then fails, eventhough I use
```python
super().build(shape)
```
in my custom *build()* to forward the call into larq. If the build call would be forward in larq, too, the *self.built* attribute would be set during the first (explicit) *build* call and the second (implicit) *build* call triggered by *tf.keras.layers.Layer.call* would be skipped.

Thank you for reading this!

lgeiger:
Thanks for spotting this. I pushed #621 to fix this.

In general `BaseQuantizer` is not a public API and it can be dangerous to use for anything other than the builtin quantizers since it overwrites
https://github.com/larq/larq/blob/094eaf1cca5017cdc643bf68bbb1301ab3f1af0f/larq/quantizers.py#L166-L168
in order to prevent the quantizer metrics from interfering with saving and loading.

I agree this is not immediately obvious (other than the fact that this class isn't part of the public docs) and can lead to confusion or hard to debug bugs. I think it would be great to add a public base class which can be used to implement quantizers.

[...]

PR #623, opened 2021-02-12
Add public base class for defining quantizers and hide BaseQuantizer

lgeiger:
This PR adds a public `lq.quantizers.Quantizer` base class which all quantizers inherit from. This can be used for typing and prevents issues like #620. This PR might be a breaking change for some people who relied on `BaseQuantizer` since I now made it private since it overwrites `non_trainable_weights` which can lead to hard to debug issues.


jneeven (approved):
Nice :rocket:

PR #627, opened 2021-02-15
Rename Quantizer type to prevent conflict with base class

lgeiger:
This type definition overwrote the base class introduced in #623 which leads to problems when trying to use `lq.quantizers.Quantizer`.

This PR renames renames the `Quantizer` type to  `QuantizerType` to prevent this conflict. Alternatively we could think about a `lq.types` module. Let's see if pytype is now happy about this.


AdamHillier (approved):
All looks good to me if typechecking passes :+1:

PR #631, opened 2021-02-19
Pass the layer's dtype to the nested quantizers

CNugteren:
This PR propagates the layer's data-type to the nested quantizers. This is important in case the Keras mixed-precision policy is used, as described [here](https://github.com/tensorflow/community/blob/master/rfcs/20200929-keras-mixed-precision.md#nesting-layers). For example, the following conversion would result in a `TypeError` before this PR, but works after these changes:

```python
import tensorflow as tf
import larq as lq

tf.keras.mixed_precision.set_global_policy("mixed_float16")

img = tf.keras.layers.Input(shape=(224, 224, 3))
x = lq.layers.QuantConv2D(3, (3, 3), input_quantizer=lq.quantizers.SteSign(clip_value=1.25))(img)
model = tf.keras.Model(img, x)

model = tf.keras.models.clone_model(
    model,
    clone_function=lambda layer: layer.__class__.from_config(
        {**layer.get_config(), "dtype": "float32"}
    )
)
```

### DIFF
--- a/larq/layers_base.py
+++ b/larq/layers_base.py
@@ -52,10 +52,10 @@ class BaseLayer(tf.keras.layers.Layer):
             return super().call(inputs)
 
     def get_config(self):
-        return {
-            **super().get_config(),
-            "input_quantizer": quantizers.serialize(self.input_quantizer),
-        }
+        quantizer_config = quantizers.serialize(self.input_quantizer)
+        if quantizer_config is not None:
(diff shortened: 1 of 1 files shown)


CNugteren:
Closing this issue after a chat with @lgeiger, since this Pr actually didn't solve the issue, instead it was hard-coding the quantizers to float32. We will think of another solution for this problem. In the meantime, if any user has such a use-case and suffers from this issue, let us know.

PR #651, opened 2021-04-16
Add DoReFa Weight Quantizer

Joschua-Conrad:
I mentioned in #620, that I use larq to implement a DoReFa weight quantizer I want to publish with this PR.

The code is added in quantizers.py together with an additional compat reference like it exists for the DoReFa activation quantizer. The new quantizer inherits from the activation quantizer, as it just adds preprocessing to the actual quantize step. In the [Paper](https://arxiv.org/pdf/1606.06160.pdf), section 2.3, one can see in formula (9), that the weight quantizer is defined by using the activation quantizer _quantizek_ together with limiting done with _tanh_ and some range-scaling operations.

I also added unit tests, which check the actual quantization and the gradient. The derivative of _tanh(x)_ is _1/cosh(x)**2_, which makes checking the gradient easy. The quantization can be checked by applying the desired preprocessing and then testing the preprocessed data just like the activation quantizer does. The tests pass.

Please check my docstring and review my comments. If I babble too much, delete the comments or tell me to do it. I could not built the docs, but I hope that my formulas look ok. I also don't know, wether saying
_The interface of this quantizer is the same like the one of DoReFa._
is ok for the docstring of if I should repeat the interface description of the DoReFa activation quantizer. And I don't know wether adding
```
plot-activation
quantizers.DoReFaKernel
```
to the docstring is ok and if the API doc quantizer figure is built automatically, but it fairly looks like it is.

And one last question: the activation quantizer is not included in _TestCommonFunctionality_, so I also did not include my own quantizer there. Is this ok?

Thank you for your time!

### DIFF
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -665,7 +665,7 @@ class DoReFaKernel(DoReFa):
         # So when all elements are zero, nothing is normed. This is achieved by
         # dividing by 1.
         dividend = tf.math.reduce_max(tf.math.abs(limited))
-        dividend = tf.where(dividend == 0., tf.ones_like(dividend), dividend)
+        dividend = tf.where(dividend == 0.0, tf.ones_like(dividend), dividend)
 
         # Need to stop the gradient here. Otherwise, for the maximum element,
         # which gives the dividend, normed is limited/limited (for this one
@@ -680,11 +680,11 @@ class DoReFaKernel(DoReFa):
 
         # Norm and scale from value range [-1,1] to [0,1]
         normed = limited / dividend
-        normed = (normed / 2.) + 0.5
+        normed = (normed / 2.0) + 0.5
 
         # Quantize and scale back to [-1,1] range
         quantized = super().call(normed)
-        quantized = (2. * quantized) - 1.
+        quantized = (2.0 * quantized) - 1.0
 
         return quantized
 
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -673,7 +673,7 @@ class DoReFaKernel(DoReFa):
         #one does the simplification y = x/x = 1. But TF does NOT do this
         #simplification when computing the gradient for the
         #normed = limited/dividend operation. As a result, this gradient becomes
-        #complex, because during the computation, "dividend" is not just a
+        #complicated, because during the computation, "dividend" is not just a
         #constant, but depends on "limited" instead. Here, tf.stop_gradient
         #is used to mark "dividend" as a constant explicitly.
         dividend = tf.stop_gradient(dividend)
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -624,7 +624,8 @@ class DoReFa(_BaseQuantizer):
         self.precision = k_bit
 
         if mode not in ("activations", "weights"):
-            raise ValueError(f"Invalid DoReFa quantizer mode {mode}.")
+            raise ValueError(f"Invalid DoReFa quantizer mode {mode}. "
+                             f"Valid values are 'activations' and 'weights'.")
         self.mode = mode
 
         super().__init__(**kwargs)
@@ -664,7 +665,8 @@ class DoReFa(_BaseQuantizer):
         elif self.mode == "weights":
             inputs = self.weight_preprocess(inputs)
         else:
-            raise ValueError(f"Invalid DoReFa quantizer mode {self.mode}.")
+            raise ValueError(f"Invalid DoReFa quantizer mode {self.mode}. "
+                             f"Valid values are 'activations' and 'weights'.")
 
         @tf.custom_gradient
         def _k_bit_with_identity_grad(x):
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -682,10 +682,7 @@ class DoReFaKernel(DoReFa):
         normed = (normed / 2.0) + 0.5
 
         # Quantize and scale back to [-1,1] range
-        quantized = super().call(normed)
-        quantized = (2.0 * quantized) - 1.0
-
-        return quantized
+        return 2.0 * super().call(normed) - 1.0
 
 
 # `DoReFa` used to be called `DoReFaQuantizer`; this alias is for
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -676,7 +676,7 @@ class DoReFaKernel(DoReFa):
         # If the dividend used for the norm operation is 0, all elements of
         # the weight tensor are 0 and divide_no_nan returns 0 for all weights.
         # So if all elements of the weight tensor are zero, nothing is normed.
-        normed = tf.math.divide_no_nan(limited, dividend) / 2.0 + 0.5
+        normed = tf.math.divide_no_nan(limited, 2.0 * dividend) + 0.5
 
         # Quantize and scale back to [-1,1] range
         return 2.0 * super().call(normed) - 1.0
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -614,7 +614,7 @@ class DoReFaKernel(DoReFa):
     with a gradient, which is contineously differentiable itself.
 
     \\[
-    w_{lim}(w) = tanh(w)
+    w_{lim}(w) = \tanh(w)
     \\]
 
     Furthermore, the weights of each layer are normed, such that the weight with
@@ -622,27 +622,27 @@ class DoReFaKernel(DoReFa):
     quantizable value. That way, the full quantizable numeric range is utilized.
 
     \\[
-    w_{norm}(w) = \frac{w}{max(|w|)}
+    w_{norm}(w) = \frac{w}{\max(|w|)}
     \\]
 
     The formulas can be found in the paper in section 2.3. Please note, that
-    the paper refers to weights being quantized on a numeric range of [-1,1], while
-    activations are quantized on the numeric range [0,1]. `DoReFa` defines the
+    the paper refers to weights being quantized on a numeric range of [-1, 1], while
+    activations are quantized on the numeric range [0, 1]. `DoReFa` defines the
     quantization function quantizek() from the paper with the correct numeric
-    range of [0,1], which is why the range needs to be adapted before applying
+    range of [0, 1], which is why the range needs to be adapted before applying
     `DoReFa` after applying the preprocessing (limit and norm).
     The hard limiting inside `DoReFa` becomes ineffective, because its input
     is already limited by the hyperbolic tangent. The full quantization
     function including the adaption of numeric ranges is
 
     \\[
-    q(w) = 2quantize_{k}(\frac{w_{norm}\left(w_{lim}\left(w\right)\right)}{2} + \frac{1}{2}) - 1
+    q(w) = 2 \, quantize_{k}(\frac{w_{norm}\left(w_{lim}\left(w\right)\right)}{2} + \frac{1}{2}) - 1
     \\]
 
     !!! warning
-        This quantizer works for weights on the range [-1,1], which matches the
+        This quantizer works for weights on the range [-1, 1], which matches the
         default setting of `constraints.weight_clip`. Do not use this quantizer
-        with a different constraint *clip_value* than the default one.
+        with a different constraint `clip_value` than the default one.
 
     ```plot-activation
     quantizers.DoReFaKernel
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -56,7 +56,7 @@ __all__ = [
     "ApproxSign",
     "DoReFa",
     "DoReFaQuantizer",
-    "DoReFaWeight",
+    "DoReFaKernel",
     "MagnitudeAwareSign",
     "NoOp",
     "NoOpQuantizer",
@@ -662,7 +662,7 @@ class DoReFaKernel(DoReFa):
         
         #Divider for max-value norm.
         #If all elements are 0., we would get a div by zero.
-        #So when all elements are zero, nothing is normed. This achieved by
+        #So when all elements are zero, nothing is normed. This is achieved by
         #dividing by 1.
         dividend = tf.reduce_max(tf.abs(limited))
         dividend = tf.where(dividend == 0., tf.ones_like(dividend), dividend)
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -678,8 +678,7 @@ class DoReFaKernel(DoReFa):
         dividend = tf.stop_gradient(dividend)
 
         # Norm and scale from value range [-1,1] to [0,1]
-        normed = limited / dividend
-        normed = (normed / 2.0) + 0.5
+        normed = limited / dividend / 2.0 + 0.5
 
         # Quantize and scale back to [-1,1] range
         return 2.0 * super().call(normed) - 1.0
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -667,6 +667,17 @@ class DoReFaKernel(DoReFa):
         dividend = tf.reduce_max(tf.abs(limited))
         dividend = tf.where(dividend == 0., tf.ones_like(dividend), dividend)
         
+        #Need to stop the gradient here. Otherwise, for the maximum element,
+        #which gives the dividend, normed is limited/limited (for this one
+        #maximum digit). The derivative of y = x/x, dy/dx is just zero, when
+        #one does the simplification y = x/x = 1. But TF does NOT do this
+        #simplification when computing the gradient for the
+        #normed = limited/dividend operation. As a result, this gradient becomes
+        #complex, because during the computation, "dividend" is not just a
+        #constant, but depends on "limited" instead. Here, tf.stop_gradient
+        #is used to mark "dividend" as a constant explicitly.
+        dividend = tf.stop_gradient(dividend)
+        
         #Norm and scale from value range [-1,1] to [0,1]
         normed = limited / dividend
         normed = (normed / 2.) + 0.5
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -601,7 +601,7 @@ class DoReFa(_BaseQuantizer):
 
     # Arguments
         k_bit: number of bits for the quantization.
-        mode: "activations" for clipping inputs on [0, 1] range or "weights" for
+        mode: `"activations"` for clipping inputs on [0, 1] range or `"weights"` for
             soft-clipping and norming weights on [-1, 1] range before applying
             quantization.
         metrics: An array of metrics to add to the layer. If `None` the metrics set in
@@ -626,7 +626,7 @@ class DoReFa(_BaseQuantizer):
         if mode not in ("activations", "weights"):
             raise ValueError(
                 f"Invalid DoReFa quantizer mode {mode}. "
-                f"Valid values are 'activations' and 'weights'."
+                "Valid values are 'activations' and 'weights'."
             )
         self.mode = mode
 
@@ -655,9 +655,7 @@ class DoReFa(_BaseQuantizer):
         # If the dividend used for the norm operation is 0, all elements of
         # the weight tensor are 0 and divide_no_nan returns 0 for all weights.
         # So if all elements of the weight tensor are zero, nothing is normed.
-        normed = tf.math.divide_no_nan(limited, 2.0 * dividend) + 0.5
-
-        return normed
+        return tf.math.divide_no_nan(limited, 2.0 * dividend) + 0.5
 
     def call(self, inputs):
         # Depending on quantizer mode (activation or weight) just clip inputs
@@ -669,7 +667,7 @@ class DoReFa(_BaseQuantizer):
         else:
             raise ValueError(
                 f"Invalid DoReFa quantizer mode {self.mode}. "
-                f"Valid values are 'activations' and 'weights'."
+                "Valid values are 'activations' and 'weights'."
             )
 
         @tf.custom_gradient
--- a/larq/quantizers.py
+++ b/larq/quantizers.py
@@ -56,6 +56,7 @@ __all__ = [
     "ApproxSign",
     "DoReFa",
     "DoReFaQuantizer",
+    "DoReFaWeight",
     "MagnitudeAwareSign",
     "NoOp",
     "NoOpQuantizer",
@@ -562,6 +563,7 @@ class DoReFa(_BaseQuantizer):
         While the DoReFa paper describes how to do quantization for both weights and
         activations, this implementation is only valid for activations, and this
         quantizer should therefore not be used as a kernel quantizer.
+        See `DoReFaKernel`.
 
     ```plot-activation
     quantizers.DoReFa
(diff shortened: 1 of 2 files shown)


jneeven:
Hi, thanks a lot for this PR! I've been wanting to look at it for a few days now, but unfortunately I'm really busy at the moment and of course I'll need to verify the correctness of your implementation with the paper etc. I'll look at this as soon as I have time do to so, but from a quick glance this looks great 🚀

Joschua-Conrad:
Take your time! Relating the formulas in the paper to my code is probably not that trivial. I hope the comments help.

larq/quantizers.py
```diff
@@ -601,9 +604,95 @@ def get_config(self):
...
+    """
+
+    def call(self, inputs):
+
+        limited = tf.math.tanh(inputs)
+
+        # Divider for max-value norm.
+        # If all elements are 0., we would get a div by zero.
+        # So when all elements are zero, nothing is normed. This is achieved by
+        # dividing by 1.
```
lgeiger, line 666: Why is this necessary? I don't see any mention of this in the paper and the linked [reference implementation](https://github.com/tensorpack/tensorpack/blob/5845ead80e3d77c89e2f1e9d5e0691324a8354cc/examples/DoReFa-Net/dorefa.py#L35) also doesn't seem to do this.

TensorFlow will return `inf` for  `x / 0`, which obviously isn't ideal, but I don't think this code would throw a divide by zero error.

larq/quantizers.py
```diff
@@ -601,9 +604,95 @@ def get_config(self):
...
+        # simplification when computing the gradient for the
+        # normed = limited/dividend operation. As a result, this gradient
+        # becomes complicated, because during the computation, "dividend" is
+        # not just a constant, but depends on "limited" instead. Here,
+        # tf.stop_gradient is used to mark "dividend" as a constant explicitly.
+        dividend = tf.stop_gradient(dividend)
+
+        # Norm and scale from value range [-1,1] to [0,1]
+        normed = limited / dividend
+        normed = (normed / 2.0) + 0.5
```
lgeiger, line 683:
```suggestion
        normed = limited / dividend - 0.5
```

larq/quantizers.py
```diff
@@ -601,9 +604,95 @@ def get_config(self):
...
+
+    def call(self, inputs):
+
+        limited = tf.math.tanh(inputs)
+
+        # Divider for max-value norm.
+        # If all elements are 0., we would get a div by zero.
+        # So when all elements are zero, nothing is normed. This is achieved by
+        # dividing by 1.
+        dividend = tf.math.reduce_max(tf.math.abs(limited))
```
lgeiger, line 667:
```suggestion
        dividend = 2 * tf.math.reduce_max(tf.math.abs(limited))
```

larq/quantizers_test.py
```diff
@@ -412,4 +462,5 @@ def test_get_kernel_quantizer_accepts_function():
 
 def test_backwards_compat_aliases():
     assert lq.quantizers.DoReFaQuantizer == lq.quantizers.DoReFa
+    assert lq.quantizers.DoReFaKernelQuantizer == lq.quantizers.DoReFaKernel
```
lgeiger, line 465:
```suggestion
```

lgeiger (commented):
Thanks again for the PR! I made a few initial comments and have some questions about the implementation.

larq/quantizers.py
Joschua-Conrad: The suggested change is missing the scaling factor. Commited the corrected suggested change in 426c64e93f029d53b8239774e262dc1ea9bb509c

larq/quantizers.py
Joschua-Conrad: Sorry when I resolved https://github.com/larq/larq/pull/651#discussion_r620329710, I did not see that you include the scaling factor into the dividend. I suggest keeping all the scaling in the last two code lines, to have the scaling and the reversal scaling all in one place. See 426c64e93f029d53b8239774e262dc1ea9bb509c .

larq/quantizers.py
```diff
@@ -601,9 +604,95 @@ def get_config(self):
         return {**super().get_config(), "k_bit": self.precision}
 
 
+@utils.register_alias("dorefa_kernel_quantizer")
+@utils.register_keras_custom_object
+class DoReFaKernel(DoReFa):
```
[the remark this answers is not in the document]
Joschua-Conrad: I chose this behavior, because a new class integrates nicely into the automatic creation of the docs and into the unittest framework, where everything is optimized for "one lass per quantizer behavior". I thought about using an instance of the activation quantizer inside the weight quantizer instead of using inheritance, but I am not sure about side effets, I would need to forward arguments to this inner instance and I would need to forward calls to get_config. Using inheritance was the cleanest solution I found.

larq/quantizers.py
Joschua-Conrad: If this would ocur in one layer, it would output _inf_ for at least one output pixel. The following batch norm layer would calculate the mean of some usual numbers and one _inf_, which is _inf_. The BN layer would apply a _- inf_ operation to all pixels to normalize the mean and the _inf_ value spreads to all pixels. So if this occurs in one spot, it ruins the entire network. Having all weights at exactly zero is anyhow a rare occasion, but having _inf_ or _-inf_ for all outputs is probably not what the user expects. So I added this check. Do you agree?

Joschua-Conrad:
Three questions are open. Please check @lgeiger

larq/quantizers.py
lgeiger: Makes sense. I looked at the code again and the only way this could happen is when all element of `limited` are exactly zero. This means we can replace `limited / dividend` with [`tf.math.divide_no_nan(limited, dividend)`](https://www.tensorflow.org/api_docs/python/tf/math/divide_no_nan) which returns `0` in cases `dividend == 0` which allows to remove the special handling on line 667 without changing the computation.

Joschua-Conrad: Done in b4a6e29db3a1a6caaa0503d8896395badc14b659

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+    r"""Instantiates a serializable k_bit kernel quantizer as in the DoReFa paper.
+
+    This quantizer is the same like the `DoReFa` quantizer, but adds a preprocessing.
+    Instead of limiting input operands (or in this case: weights) using a hard
+    limiter, a tangens hyperbolicus is applied to achieve a softer limiting
+    with a gradient, which is contineously differentiable itself.
+
+    \\[
+    w_{lim}(w) = tanh(w)
```
lgeiger, line 617:
```suggestion
    w_{lim}(w) = \tanh(w)
```

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+    \\]
+
+    Furthermore, the weights of each layer are normed, such that the weight with
+    the largest magnitude gets the largest or smallest (depending on its sign)
+    quantizable value. That way, the full quantizable numeric range is utilized.
+
+    \\[
+    w_{norm}(w) = \frac{w}{max(|w|)}
```
lgeiger, line 625:
```suggestion
    w_{norm}(w) = \frac{w}{\max(|w|)}
```

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+    \\]
+
+    The formulas can be found in the paper in section 2.3. Please note, that
+    the paper refers to weights being quantized on a numeric range of [-1,1], while
```
lgeiger, line 629:
```suggestion
    the paper refers to weights being quantized on a numeric range of [-1, 1], while
```

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+    activations are quantized on the numeric range [0,1]. `DoReFa` defines the
```
lgeiger, line 630:
```suggestion
    activations are quantized on the numeric range [0, 1]. `DoReFa` defines the
```

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+    quantization function quantizek() from the paper with the correct numeric
+    range of [0,1], which is why the range needs to be adapted before applying
```
lgeiger, line 632:
```suggestion
    range of [0, 1], which is why the range needs to be adapted before applying
```

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+    `DoReFa` after applying the preprocessing (limit and norm).
+    The hard limiting inside `DoReFa` becomes ineffective, because its input
+    is already limited by the hyperbolic tangent. The full quantization
+    function including the adaption of numeric ranges is
+
+    \\[
+    q(w) = 2quantize_{k}(\frac{w_{norm}\left(w_{lim}\left(w\right)\right)}{2} + \frac{1}{2}) - 1
```
lgeiger, line 639:
```suggestion
    q(w) = 2 \, quantize_{k}(\frac{w_{norm}\left(w_{lim}\left(w\right)\right)}{2} + \frac{1}{2}) - 1
```

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+    \\]
+
+    !!! warning
+        This quantizer works for weights on the range [-1,1], which matches the
```
lgeiger, line 643:
```suggestion
        This quantizer works for weights on the range [-1, 1], which matches the
```

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+        default setting of `constraints.weight_clip`. Do not use this quantizer
+        with a different constraint *clip_value* than the default one.
```
lgeiger, line 645:
```suggestion
        with a different constraint `clip_value` than the default one.
```

lgeiger (commented):
I added a few minor suggestions to improve docs rendering.

lgeiger (approved):
This is great! Thanks a lot for the PR 🚀 

@larq/core Could we get another pair of eyes to review this before we merge?

jneeven:
> @larq/core Could we get another pair of eyes to review this before we merge?

I hope to look at this early next week!

larq/quantizers.py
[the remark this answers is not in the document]
jneeven: For cleanliness of the code I'd prefer to have just a single quantizer class (e.g. like the preliminary version I wrote [here](https://github.com/larq/larq/issues/487#issuecomment-630748936)), but the docs are a good point. I think I would slightly prefer having a base `DoReFa` with `mode="activations"` and then just create a `DoReFaKernel` wrapper that sets `mode="weights"`. No strong opinions though.

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
 
+@utils.register_alias("dorefa_kernel_quantizer")
+@utils.register_keras_custom_object
+class DoReFaKernel(DoReFa):
+    r"""Instantiates a serializable k_bit kernel quantizer as in the DoReFa paper.
+
+    This quantizer is the same like the `DoReFa` quantizer, but adds a preprocessing.
+    Instead of limiting input operands (or in this case: weights) using a hard
+    limiter, a tangens hyperbolicus is applied to achieve a softer limiting
+    with a gradient, which is contineously differentiable itself.
```
jneeven, line 614:
```suggestion
    with a gradient, which is continuously differentiable itself.
```

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+    activations are quantized on the numeric range [0, 1]. `DoReFa` defines the
+    quantization function quantizek() from the paper with the correct numeric
+    range of [0, 1], which is why the range needs to be adapted before applying
+    `DoReFa` after applying the preprocessing (limit and norm).
+    The hard limiting inside `DoReFa` becomes ineffective, because its input
+    is already limited by the hyperbolic tangent. The full quantization
+    function including the adaption of numeric ranges is
+
+    \\[
+    q(w) = 2 \, quantize_{k}(\frac{w_{norm}\left(w_{lim}\left(w\right)\right)}{2} + \frac{1}{2}) - 1
```
jneeven, line 639: @lgeiger  can we get a render preview somewhere?

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+    The interface of this quantizer is the same like the one of `DoReFa`.
+
+    # References
+        - [DoReFa-Net: Training Low Bitwidth Convolutional Neural Networks with Low
+            Bitwidth Gradients](https://arxiv.org/abs/1606.06160)
+    """
+
+    def call(self, inputs):
+
+        limited = tf.math.tanh(inputs)
```
jneeven, line 660:
```suggestion
        limited = tf.math.tanh(inputs)
```

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
...
+        # becomes complicated, because during the computation, "dividend" is
+        # not just a constant, but depends on "limited" instead. Here,
+        # tf.stop_gradient is used to mark "dividend" as a constant explicitly.
+        dividend = tf.stop_gradient(dividend)
+
+        # Norm and then scale from value range [-1,1] to [0,1].
+        # If the dividend used for the norm operation is 0, all elements of
+        # the weight tensor are 0 and divide_no_nan returns 0 for all weights.
+        # So if all elements of the weight tensor are zero, nothing is normed.
+        normed = tf.math.divide_no_nan(limited, dividend) / 2.0 + 0.5
```
jneeven, line 680:
```suggestion
        normed = tf.math.divide_no_nan(limited, 2.0 * dividend) + 0.5
```
This keeps the notation closer to the formula in the paper, if you don't mind

larq/quantizers.py
```diff
@@ -601,6 +603,86 @@ def get_config(self):
         return {**super().get_config(), "k_bit": self.precision}
 
 
+@utils.register_alias("dorefa_kernel_quantizer")
+@utils.register_keras_custom_object
+class DoReFaKernel(DoReFa):
+    r"""Instantiates a serializable k_bit kernel quantizer as in the DoReFa paper.
```
jneeven, line 609: When `k=1`, the DoReFa weight quantizer equals `SteSign`. Is this currently handled correctly?

larq/quantizers_test.py
```diff
@@ -234,6 +234,33 @@ def test_dorefa_quantize(self):
...
+        # to be done here like in the weight quantizer when comparing digits
+        # of the quantized tensor to golden, quantized values. The golden,
+        # quantized value "i / n" is on the [0,1] range.
+        for i in range(n + 1):
+            np.testing.assert_allclose(
+                result[
+                    (preprocessed > (2 * i - 1) / (2 * n))
+                    & (preprocessed < (2 * i + 1) / (2 * n))
+                ],
+                2 * (i / n) - 1.0,
```
jneeven, line 261: @lgeiger I see this computation was already in the original dorefa unit test, what's going on here? For some reason the index is some complicated `AND` operation?

larq/quantizers_test.py
```diff
@@ -339,6 +366,28 @@ def ste_grad(x):
...
+            # quantizek function is supposed to be 1 everywhere, because it
+            # is used on its linear region only. tanh does all the limiting.
+            dividend = np.amax(np.abs(np.tanh(x)))
+            return 1 / (np.cosh(x) ** 2.0) / dividend
+
+        x = testing_utils.generate_real_values_with_zeros(shape=(8, 3, 3, 16))
+        tf_x = tf.Variable(x)
+        with tf.GradientTape() as tape:
+            activation = lq.quantizers.DoReFaKernel(2)(tf_x)
+        grad = tape.gradient(activation, tf_x)
```
jneeven, line 388: Can you add a case for `k=1` as well?

jneeven (commented):
Looks good! Thanks a lot for this. I just have some questions and minor remarks.

larq/quantizers.py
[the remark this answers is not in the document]
Joschua-Conrad: This decision affects building the docs and testing, where I do not know the larq project as good as others do. Please agree on one approach and I'll change (or not change) the code accordingly. I took the subclass-way, because it leaves a clear border between activation and weight quantizer without the need for forwarding any calls. Instead using a new keyword for the activation quantizer and some kind of a wrapper for creating the weight quantizer skips usage of inheritance and possible pitfalls.

larq/quantizers.py
Joschua-Conrad: It is, because the weight quantizer only adds a preprocessor for the activation quantizer.

larq/quantizers_test.py
Joschua-Conrad: Here, ``result`` and ``preprocessed`` have exactly the same shape. One could feed ``preprocessed`` into the activation quantizer and would get exactly ``result``. 

``
(preprocessed > (2 * i - 1) / (2 * n))
& (preprocessed < (2 * i + 1) / (2 * n))
``

creates a boolean index matrix, which marks the position of the values, which fall into the current quantization step bound ``i``. So after applying this matrix to the quantized ``result``, all the returned elements need to have the same value, because they all where inside one quantization step bound in ``preprocessed`` and were quantized to the exact same value.

larq/quantizers_test.py
Joschua-Conrad: Thanky for that remark, this is indeed missing!

Done in 82df95fed124151b278096e81620a6253283737f

This also demonstrates validity of answer in https://github.com/larq/larq/pull/651#discussion_r625002136

Had to increase tolerance of numpy's ``assert_allclose`` by a factor of 10. But this is due to numeric instability. The relative tolerance is 1e-6. The relative tolerance of an 8 bit quantization is 1/256 = 3.9e-3, so actual quantization errors will still be detected.

Joschua-Conrad:
@jneeven Thank your for your review. I added testcases for more than just a 2 bit quantization and accepted your suggestions.

I'm awaiting the following decisions:

- Is the k_bit testcase ok?
- Do we add the different k_bit testcases also to the activation quantizer?
- Do I need to create a ``mode`` argument for the activation quantizer and restructure the implementation of the weight quantizer?

larq/quantizers_test.py
jneeven: Looks great, thanks!

larq/quantizers_test.py
jneeven: Makes sense, thanks!

larq/quantizers.py
lgeiger: This is done in https://github.com/larq/docs and unfortunately we don't have an automated way to render a preview. However, I manually checked and it renders correctly.

jneeven (approved):
> Is the k_bit testcase ok?

Looks good to me!

> Do we add the different k_bit testcases also to the activation quantizer?

Would be nice! But we can also do that in a separate PR.

> Do I need to create a mode argument for the activation quantizer and restructure the implementation of the weight quantizer?

This works, so let's leave it like this :+1:

Joschua-Conrad:
Thank you for reviewing!

> > Do we add the different k_bit testcases also to the activation quantizer?
> 
> Would be nice! But we can also do that in a separate PR.

I will create a new PR as soon as this one is merged and I will do so using the updated master branch. If we then find some issue with the multibit testing, we maybe need to edit activation and weight quantizer testing at the same time so I want the weight quantizer multibit testing to be already there.

larq/quantizers.py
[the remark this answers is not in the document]
lgeiger: Thinking about this again, we can easily change the docs generation to add autogenerated plots also for quantizers with specific arguments (I am happy to take this on in a follow up PR since it requires changes in some not well documented code paths).

Following the argument from @jneeven above I think I now agree that using a single `DoReFa` class with a `mode` argument that defaults to `"activations"` is probably the more consistent approach within our current API. This might come up again in the future where different quantizers might need different behaviour depending on whether they are used for weights or activations. An example for this would be  int8 quantizers that require different per-channel setting depending on whether they are used for weights or activations. Adding separate classes for each of these would probably increase the API surface too much in my opinion.
I think I am also against exposing a separate `DoReFaKernel` wrapper in this case so this PR would reduce to adding support for a `mode` argument to the existing `DoReFa` class. Would this work for your use case?

Sorry for the back and forth but since this changes the public API I think we want to get it right so we don't have to make any breaking changes later on.

lgeiger (changes requested):

lgeiger:
@Joschua-Conrad Thanks for the fast reply and sorry for the back and forth. Can you take a look at my comment above regarding the public API?

Other than that this looks great and I hope we can merge this soon and make a new release.

Joschua-Conrad:
> @Joschua-Conrad Thanks for the fast reply and sorry for the back and forth. Can you take a look at my comment above regarding the public API?
> 
> Other than that this looks great and I hope we can merge this soon and make a new release.

No prob. Please check my roadmap for the upcoming changes.

@lgeiger @jneeven

larq/quantizers.py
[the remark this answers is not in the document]
lgeiger: This sounds great!

larq/quantizers_test.py
```diff
@@ -400,7 +382,6 @@ def tanh_grad(x):
         ("magnitude_aware_sign", lq.quantizers.MagnitudeAwareSign),
         ("ste_tern", lq.quantizers.SteTern),
         ("dorefa_quantizer", lq.quantizers.DoReFa),
-        ("dorefa_kernel_quantizer", lq.quantizers.DoReFaKernel),
```
lgeiger: If you still want to test this, I think we can use something like:
```python
("dorefa_kernel_quantizer", functools.partial(lq.quantizers.DoReFa, mode="weights")),
```
Otherwise, I think it is also fine to not add it here.

larq/quantizers.py
```diff
@@ -558,47 +558,129 @@ class DoReFa(_BaseQuantizer):
...
+            quantization.
         metrics: An array of metrics to add to the layer. If `None` the metrics set in
             `larq.context.metrics_scope` are used. Currently only the `flip_ratio`
             metric is available.
 
     # Returns
         Quantization function
 
+    # Raises
+        ValueError for bad value of *mode*.
```
lgeiger, line 615:
```suggestion
        ValueError for bad value of `mode`.
```

larq/quantizers.py
```diff
@@ -558,47 +558,129 @@ class DoReFa(_BaseQuantizer):
...
         - [DoReFa-Net: Training Low Bitwidth Convolutional Neural Networks with Low
             Bitwidth Gradients](https://arxiv.org/abs/1606.06160)
     """
     precision = None
 
-    def __init__(self, k_bit: int = 2, **kwargs):
+    def __init__(self, k_bit: int = 2, mode: str = "activations", **kwargs):
         self.precision = k_bit
+
+        if mode not in ("activations", "kernel"):
```
lgeiger, line 626: I think using `mode in ("activations", "weights")` is more consistent with the naming Keras uses, since kernel can have multiple meanings depending on the context.

larq/quantizers.py
```diff
@@ -558,47 +558,129 @@ class DoReFa(_BaseQuantizer):
...
+            raise ValueError(f"Invalid DoReFa quantizer mode {mode}.")
```
lgeiger, line 627: Would be great if the error message would also list the available modes. This can be very useful if there is a small typo in the argument that might be easy to overlook.

larq/quantizers.py
```diff
@@ -558,47 +558,129 @@ class DoReFa(_BaseQuantizer):
...
+        self.mode = str(mode)
```
lgeiger, line 628: If mode is not already a string, the line above would throw.
```suggestion
        self.mode = mode
```

lgeiger (changes requested):
This is great! some more final comment, before this is good to go 🚀

larq/quantizers.py
Joschua-Conrad: Done in 363c60cccdaaf69924173ffa8f74a899d71d8c38

larq/quantizers.py
Joschua-Conrad: List available modes: ef76f89b1e170833b3fd4a429960a984c6decca7
Test invalid mode exceptions: b3fbc421cf4bdd68d546bfc438fa310d70fee217

larq/quantizers_test.py
Joschua-Conrad: Added testcase in b4bfd473f3ca7878dd10b6122a9d386404f73d89

Added
```
("dorefa_quantizer", functools.partial(lq.quantizers.DoReFa, mode="weights")),
```

because the quantizer string needs to be a valid one, because the test routine passes the quantizer name for some layers, which try to retrieve a valid class for the name string.

Please check!

Joschua-Conrad:
@lgeiger Done. Only one point remains open.

I also added the black conversions and the exception unittests, such that all CI runs should pass now.

larq/quantizers.py
```diff
@@ -558,47 +558,135 @@ class DoReFa(_BaseQuantizer):
...
+        default setting of `constraints.weight_clip`. Do not use this quantizer
+        with a different constraint `clip_value` than the default one.
 
     ```plot-activation
     quantizers.DoReFa
     ```
 
     # Arguments
         k_bit: number of bits for the quantization.
+        mode: "activations" for clipping inputs on [0, 1] range or "weights" for
```
lgeiger, line 604:
```suggestion
        mode: `"activations"` for clipping inputs on [0, 1] range or `"weights"` for
```

larq/quantizers.py
```diff
@@ -558,47 +558,135 @@ class DoReFa(_BaseQuantizer):
...
     precision = None
 
-    def __init__(self, k_bit: int = 2, **kwargs):
+    def __init__(self, k_bit: int = 2, mode: str = "activations", **kwargs):
         self.precision = k_bit
+
+        if mode not in ("activations", "weights"):
+            raise ValueError(
+                f"Invalid DoReFa quantizer mode {mode}. "
+                f"Valid values are 'activations' and 'weights'."
```
lgeiger, line 629:
```suggestion
                "Valid values are 'activations' and 'weights'."
```

larq/quantizers.py
```diff
@@ -558,47 +558,135 @@ class DoReFa(_BaseQuantizer):
...
+        dividend = tf.stop_gradient(dividend)
+
+        # Norm and then scale from value range [-1,1] to [0,1] (the range
+        # expected by the core quantization operation).
+        # If the dividend used for the norm operation is 0, all elements of
+        # the weight tensor are 0 and divide_no_nan returns 0 for all weights.
+        # So if all elements of the weight tensor are zero, nothing is normed.
+        normed = tf.math.divide_no_nan(limited, 2.0 * dividend) + 0.5
+
+        return normed
```
lgeiger, line 660:
```suggestion
        return tf.math.divide_no_nan(limited, 2.0 * dividend) + 0.5
```

larq/quantizers.py
```diff
@@ -558,47 +558,135 @@ class DoReFa(_BaseQuantizer):
...
+        # Depending on quantizer mode (activation or weight) just clip inputs
+        # on [0, 1] range or use weight preprocessing method.
+        if self.mode == "activations":
+            inputs = tf.clip_by_value(inputs, 0.0, 1.0)
+        elif self.mode == "weights":
+            inputs = self.weight_preprocess(inputs)
+        else:
+            raise ValueError(
+                f"Invalid DoReFa quantizer mode {self.mode}. "
+                f"Valid values are 'activations' and 'weights'."
```
lgeiger, line 672:
```suggestion
                "Valid values are 'activations' and 'weights'."
```

larq/quantizers_test.py
```diff
@@ -325,19 +345,39 @@ def test_magnitude_aware_sign_grad(self):
...
             if x <= 1 and x >= 0:
                 return 1.0
             return 0.0
 
+        # For other tests, the golden gradient is defined using python
+        # expressions and "converted" to a numpy function using np.vectorize.
+        # But the weight quantizer does a max-operation over the full
+        # quantized tensor, so np.vectorize, which causes the gradient to be
+        # called element-wise, needs to be skipped!
+        # @np.vectorize
```
lgeiger, line 361: For the other gradient test we only use `np.vectorize`, because we require an `if/else` statement which is easier to read that way. Since this is not the case here, the native numpy usage is actually preferred. So we can remove this comment.
```suggestion
```

larq/quantizers_test.py
```diff
@@ -325,19 +345,39 @@ def test_magnitude_aware_sign_grad(self):
...
+            # scaling operations cancel each other and the gradient of the
+            # quantizek function is supposed to be 1 everywhere, because it
+            # is used on its linear region only. tanh does all the limiting.
+            dividend = np.amax(np.abs(np.tanh(x)))
+            return 1 / (np.cosh(x) ** 2.0) / dividend
+
+        if mode == "activations":
+            expected_gradient = ste_grad
+        else:
+            expected_gradient = tanh_grad
```
lgeiger, line 373:
```suggestion
        expected_gradient = ste_grad if mode == "activations" else tanh_grad
```

lgeiger (approved):
Thanks for sticking with this PR through the long review. I only have a few minor suggestions that can be directly applied, other than that this is now ready to merge.

Thanks again, for your patience.

Joschua-Conrad:
Thank you for explaining in a nice and patient way what you wanted to be changed! I learned quite a lot about how to do things right in a review process.

lgeiger:
> Thank you for explaining in a nice and patient way what you wanted to be changed! I learned quite a lot about how to do things right in a review process.

Thanks for the feedback, looking forward to the next PR :)

lgeiger:
The changes are now released as part of larq 0.12 and the docs are available at https://docs.larq.dev/larq/api/quantizers/#dorefa

PR #677, opened 2021-06-04
Simplify lq.math implementations

lgeiger:
This PR simplifies the implementations of `lq.math` to make them more readable. Currently this would break Compute Engine, since we rely on the implementation details there. We could either fix this using `@tf.function(experimental_implements=...)` which would require us dropping support for older TensorFlow version, or we could update the patterns in Compute Engine to handle this implementation properly which shouldn't be hard.

I briefly evaluated the performance impact of this change on a T4 GPU in [this notebook](https://colab.research.google.com/drive/16YcyqFmVNaFfXhlqItUYsoL2HF8iX8-I?usp=sharing) (although only on one input size). This shows performance improvements for some cases and performance regressions for others, although I am not sure if this would be noticeable in a real world model. I am happy to run benchmarks if you think this change might impact full model training performance.

This PR is still WIP for now to explore this further.


Tombana:
> We could either fix this using `@tf.function(experimental_implements=...)` which would require us dropping support for older TensorFlow version, or we could update the patterns in Compute Engine ..

Minor comment: `@tf.function(experimental_implements=...)` doesn't make it convert, it only adds a tag that the converter can read, you'd still need to add the pattern to compute engine, so you might as well make it a proper pattern.

lgeiger:
> Minor comment: @tf.function(experimental_implements=...) doesn't make it convert, it only adds a tag that the converter can read, you'd still need to add the pattern to compute engine, so you might as well make it a proper pattern.

Yes, you are right. The idea behind it was that we then would be implementation independent, but I agree a proper pattern is much easier and we can still keep it backwards compatible.

AdamHillier:
I agree that matching directly against the pattern in LCE would be good. If it doesn't cause any issues, it might be nice to still keep the `experimental_implements` so that if we want to change the converter implementation to use it at some point in the future then that 'tag' will already be there in existing versions of Larq.

How do you think we should manage the scenario where somebody uses a new version of Larq but an old version of the converter? I guess on init we could try and detect if LCE is installed, detect the version, and print some kind of warning message if it's `<= 0.6`?

lgeiger:
> How do you think we should manage the scenario where somebody uses a new version of Larq but an old version of the converter? I guess on init we could try and detect if LCE is installed, detect the version, and print some kind of warning message if it's <= 0.6?

Yes, I think that should work well. That's how [TensorFlow Addons handles it as well](https://github.com/tensorflow/addons/blob/e83e71cf07f65773d0f3ba02b6de66ec3b190db7/tensorflow_addons/__init__.py#L18).

lgeiger:
I added a LCE version check to `__init__` in e24265d84b909f20cc844881f645e98def8cc2c5 now that LCE 0.6.1 has support for this.

I'd be happy to move this check to `setup.py` though in case warning during import is too verbose.

larq/math.py
```diff
@@ -22,7 +22,8 @@ def sign(x):
     # Returns
         A Tensor with same type as `x`.
     """
-    return tf.sign(tf.sign(x) + 0.1)
+    ones = tf.ones_like(x)
+    return tf.where(x >= 0, ones, -ones)
```
AdamHillier, line 26: I'm not sure if there's any possible benefit to doing so in terms of speed/memory usage, but LCE would also correctly convert `tf.where(x >= 0, 1.0, -1.0)` if we just rely on implicit broadcasting.

larq/utils.py
```diff
@@ -68,3 +70,17 @@ def patch_object(object, name, value):
...
+        import larq_compute_engine as lce  # pytype: disable=import-error
+
+        lce_version = getattr(lce, "__version__", "0.0.0")
+        if version.parse(lce_version) < version.parse("0.6.1"):
+            warnings.warn(
+                f"larq-compute-engine={lce_version} is not supported by this version of larq.\n"
+                "Please upgrade larq-compute-engine to v0.6.1 or newer."
+            )
+    except ImportError:
+        pass
```
AdamHillier, line 86: How do you feel about instead doing the version check of LCE using `importlib.metadata`, which would mean we wouldn't have to import LCE? Like we do for getting the Larq version itself:
https://github.com/larq/larq/blob/4ae5cc9e8f49a8db844e6e0c84d52c23eac10fee/larq/__init__.py#L15-L21

I only ask because at the moment Larq imports very quickly, whereas LCE imports very slowly (as slowly as TF does).

Alternatively we could do the check in `setup.py` as you say, though I think the problem there is that that means a warning would not be shown if first Larq is installed, and then LCE.

lgeiger: Good idea! I thought the main slowness of the LCE import comes from TF itself, but it's good to just use `importlib.metadata` anyway since it's super fast.

lgeiger: Good idea! I thought the main slowness of the LCE import comes from TF itself, but it's good to just use `importlib.metadata` anyway since it's super fast.

larq/math.py
lgeiger: I briefly benchmarked this some time ago on a T4 GPU in [this notebook](https://colab.research.google.com/drive/16YcyqFmVNaFfXhlqItUYsoL2HF8iX8-I?usp=sharing) (although only on one input size) and it looked like explicit broadcasting using `ones_like` is faster when running with default TF kernels (especially in graph mode). If the model is XLA compiled this is optimised away and there is no performance difference. So I think using explicit broadcasting here ensures we're on the save side with respect to performance.

This is correctly supported by LCE as well, right?

AdamHillier: Yep it is, this will definitely work too :+1:

AdamHillier (approved):
Awesome :)

I guess we should run a quick sanity check model conversion with LCE 0.6.1 before merging, but this looks great.

lgeiger:
> I guess we should run a quick sanity check model conversion with LCE 0.6.1 before merging, but this looks great.

I will run a sanity check later today or tomorrow to be save!

simonmaurer:
@lgeiger @AdamHillier: just as a suggestion since I'm dealing with `larq.quantizers.SteSign` and `larq.quantizers.SteHeaviside` model conversions.
for `larq.math.sign` and `larq.math.heaviside` to use:
```
tf.where(x >= 0., 1., -1.)
```
instead of
```
tf.where(x >= 0, tf.ones_like(x), -tf.ones_like(x))
```
This would result in fewer ops.

lgeiger:
@simonmaurer Did you do any performance profiling of you proposed solution?

One problem with using `tf.where(x >= 0., 1., -1.)` is that this might change the datatype (e.g. it will always return float32, even if x is float16 or int8). This means we would need to use something like `tf.where(x >= 0, tf.constant(1, dtype=x.dtype), tf.constant(-1, dtype=x.dtype))` or do some casting which wouldn't reduce the number of ops.
I briefly profiled some of the options on a T4 GPU in [this notebook](https://colab.research.google.com/drive/16YcyqFmVNaFfXhlqItUYsoL2HF8iX8-I?usp=sharing), but please let me know if I am missing something here.

simonmaurer:
@lgeiger ah I see, this is indeed good reasoning. I didn't do any performance tests yet, my assumption would be this is negligible for one op to get the shape.
I implicitly assumed you are working with `tf.float32` tensors since this is the data type that `LceDequantize` outputs in the end.
but now thinking about it again, the `LceDequantize` also handles dequantization to `tf.int8`, otherwise there would be no need for the additional `Dequantize` op at the model output. so good point.

PR #708, opened 2021-10-06
Properly pass `groups` and `dilation_rate` parameters to convolutional layers

lgeiger:
This PR doesn't add any functionality since the parameters were already handled via `kwargs`, but it makes sure that our unittests pass and improves the documentation of our layers.

I am not sure why this only made our CI fail very recently, but so I added TF 2.6 and Python 3.9 to our CI matrix as well, just to be save.

This pulls in https://github.com/tensorflow/tensorflow/commit/5f2e159a58d1ef3414b2c34339266449574d8f94 and https://github.com/tensorflow/tensorflow/commit/46c79db67db33d7a14e124fc0b5cf23e0910d039

### DIFF
--- a/larq/layers.py
+++ b/larq/layers.py
@@ -212,7 +212,9 @@ class QuantConv1D(QuantizerBase, QuantizerBaseConv, tf.keras.layers.Conv1D):
             if version.parse(tf.__version__) >= version.parse("2.3"):
                 kwargs = {**kwargs, "groups": groups}
             else:
-                raise ValueError("`groups` != 1 requires TensorFlow version 2.3 or newer.")
+                raise ValueError(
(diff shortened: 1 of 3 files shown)


larq/layers.py
```diff
@@ -201,6 +208,11 @@ def __init__(
         bias_constraint=None,
         **kwargs,
     ):
+        if groups != 1:
+            if version.parse(tf.__version__) >= version.parse("2.3"):
+                kwargs = {**kwargs, "groups": groups}
+            else:
+                raise ValueError("`groups` != 1 requires TensorFlow version 2.3 or newer.")
```
lgeiger, line 215: I don't really like the added boilerplate here, but I think it is nicer to have the groups parameter available in the signature and not hidden in the `kwargs`.

PR #718, opened 2021-10-29
Support nested models in ModelProfile

jneeven:
Prevents the number of MACs being 0 and some other stats being slightly off.


lgeiger (changes requested):
Thanks for the fix! Can you add a [test case](https://github.com/larq/larq/blob/main/larq/models_test.py) for this?

PR #719, opened 2021-10-29
Fix LayerProfile class check with SavedModels

jneeven:
```python
(
    isinstance(layer, tensorflow.python.keras.saving.saved_model.load.QuantConv2D) != 
    isinstance(layer, larq.layers.QuantConv2D)
)
```

Unfortunately there is no way to check whether the `RevivedLayer` with name `QuantConv2D` was originally a `larq` layer and not some custom layer with the same name, but the only situation in which that'd break is if you not only subclass a `larq` layer with an identical name, but also change whether it has MACs, which seems extremely unlikely.

### DIFF
--- a/larq/models_test.py
+++ b/larq/models_test.py
@@ -1,12 +1,11 @@
-from tempfile import TemporaryDirectory
-
 import numpy as np
 import pytest
 import tensorflow as tf
 from packaging import version
+from tensorflow.keras.utils import get_custom_objects
 
 import larq as lq
-from larq.models import ModelProfile
+from larq.models import ModelProfile, is_of_layer_type, mac_containing_layers
 
 
 class ToyModel(tf.keras.Model):
@@ -246,13 +245,14 @@ def test_subclass_model_summary(snapshot, capsys):
     snapshot.assert_match(captured.out)
 
 
-def test_saved_model_summary(capsys):
+def test_saved_model_summary(tmpdir):
     model = tf.keras.models.Sequential(
         [
             lq.layers.QuantConv2D(
                 filters=32,
                 kernel_size=(3, 3),
                 kernel_quantizer="ste_sign",
+                input_quantizer="ste_sign",
                 input_shape=(64, 64, 1),
                 padding="same",
             ),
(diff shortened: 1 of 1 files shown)


larq/models.py
```diff
@@ -36,6 +38,35 @@
...
+    if isinstance(layer, op_count_supported_layer_types):
+        return True
+
+    if version.parse(tf.__version__) >= version.parse("2.0"):
+        from tensorflow.python.keras.saving.saved_model.load import RevivedLayer
+
+        names = {cls.__name__ for cls in op_count_supported_layer_types}
+        return isinstance(layer, RevivedLayer) and layer.__class__.__name__ in names
+
+    return False
```
CNugteren, line 67: These functions are almost the same. If I am right you can just add one variable as argument and replace both functions by one, e.g. something like:
```suggestion
def is_of_layer_type(layer: tf.keras.layers.Layer, layer_list: List) -> bool:
    # Check if the layer is either one of the above, or a SavedModel version of it.
    if isinstance(layer, layer_list):
        return True

    if version.parse(tf.__version__) >= version.parse("2.0"):
        from tensorflow.python.keras.saving.saved_model.load import RevivedLayer

        names = {cls.__name__ for cls in layer_list}
        return isinstance(layer, RevivedLayer) and layer.__class__.__name__ in names

    return False
```

larq/models.py
```diff
@@ -4,6 +4,7 @@
 
 import numpy as np
 import tensorflow as tf
+from packaging import version
```
lgeiger, line 7: This is currently only a dev dependency, we'd need to add this to the proper install requirements: https://github.com/larq/larq/blob/43091316fff791b03f0eceabb4a75dc088ef41a6/setup.py#L34

lgeiger (changes requested):
Cool! Can you also add a unittest to make sure we don't run into this issue again?

larq/models_test.py
```diff
@@ -244,6 +246,33 @@ def test_subclass_model_summary(snapshot, capsys):
...
+            ),
+            tf.keras.layers.GlobalAvgPool2D(),
+            tf.keras.layers.Dense(10, activation="softmax"),
+        ]
+    )
+    lq.models.summary(model)
+    summary = capsys.readouterr().out
+
+    # Save and reload
+    with TemporaryDirectory() as dir:
```
lgeiger, line 267: You can use the [pytest `tempdir` fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture) directly, instead of creating it manually.

lgeiger (changes requested):

larq/models_test.py
CNugteren: Even better to use the more modern `tmp_path: Path` from pytest, right?

CNugteren (approved):

jneeven:
@lgeiger The test was a good call, it turns out there is a more serious issue at play here as well. I've pushed an unfinished test that prints the weightprofiles of the layer before and after SavedModel loading, and they're different:
```python
# op_profiles, [w.bitwidth for w in profile.weight_profiles], mac_containing_layer, input_precision
[OperationProfile(n=1179648, precision=1, op_type='mac')] [1, 32] True 1
[OperationProfile(n=1179648, precision=32, op_type='mac')] [32, 32] True 1
```

At this point this is a bit beyond the parts of Larq I'm familiar with, so I'll have to postpone fixing this for now. I suspect the weight precision somehow isn't part of the layer config or something like that...

jneeven:
To expand on the above, the problem is as follows:
```python
from tempfile import TemporaryDirectory

import tensorflow as tf
from tensorflow.python.keras.utils.generic_utils import get_custom_objects

import larq as lq

model = tf.keras.models.Sequential(
    [
        lq.layers.QuantConv2D(
            filters=32,
            kernel_size=(3, 3),
            kernel_quantizer="ste_sign",
            input_quantizer="ste_sign",
            input_shape=(64, 64, 1),
            padding="same",
        )
    ]
)

# Save and reload
with TemporaryDirectory() as dir:
    model.save(dir)
    del get_custom_objects()["QuantConv2D"]
    loaded_model = tf.keras.models.load_model(dir, compile=False)

# Pre-save
print(type(model.layers[0].weights[0]))
print(model.layers[0].weights[0].precision)


# Loaded model
print(type(loaded_model.layers[0].weights[0]))
print(loaded_model.layers[0].weights[0].precision)
```

```
<class 'larq.quantized_variable.QuantizedVariable'>
1
<class 'tensorflow.python.ops.resource_variable_ops.UninitializedVariable'>
Traceback (most recent call last):
  File "/mnt/windows/share/Plumerai/larq/test.py", line 34, in <module>
    print(loaded_model.layers[0].weights[0].precision)
AttributeError: 'UninitializedVariable' object has no attribute 'precision'
```

PR #725, opened 2021-11-05
:arrow_up: TensorFlow on CI

lgeiger:


.github/workflows/unittest.yml
```diff
@@ -11,16 +11,18 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tf-version: [1.14.0, 1.15.5, 2.0.4, 2.1.4, 2.2.3, 2.3.3, 2.4.2, 2.5.0, 2.6.0, 2.7.0rc0]
+        tf-version: [1.14.0, 1.15.5, 2.0.4, 2.1.4, 2.2.3, 2.3.4, 2.4.4, 2.5.2, 2.6.2, 2.7.0]
         python-version: [3.7]
         include:
-          - tf-version: 2.4.2
+          - tf-version: 2.4.4
```
CNugteren, line 17: Is it on purpose that TF 2.4 appears twice here, or should one of them be 2.3.4?

lgeiger: It is on purpose, since we are testing against two different Python versions here. But we could also extend it to test against all possible combinations of Python and TensorFlow versions.

CNugteren (approved):

PR #735, opened 2021-12-02
Do not map variables according to names in `CaseOptimizer`

lgeiger:
Previously we assigned variables to their respective optimizers via their names. This works well in most cases, however we've run into an edge case where this breaks when retraining some networks that were reloaded from a saved model.

This PR changes the mapping to use `_var_key` as a unique identifier which is also used to [create slot variables (e.g. momentum) in the upstream implementation](https://github.com/keras-team/keras/blob/a8606fd45b760cce3e65727e9d62cae796c45930/keras/optimizer_v2/optimizer_v2.py#L882). Unfortunately I am unable to reproduce this failure in a dedicated unittest, but I tested this using mixed precision on a multi GPU machine.

### DIFF
--- a/larq/optimizers.py
+++ b/larq/optimizers.py
@@ -46,6 +46,24 @@ from larq import utils
 __all__ = ["Bop", "CaseOptimizer"]
 
 
+def _var_key(var):
+    """Key for representing a primary variable, for looking up slots.
+    In graph mode the name is derived from the var shared name.
+    In eager mode the name is derived from the var unique id.
+    If distribution strategy exists, get the primary variable first.
+    Args:
+      var: the variable.
+    Returns:
+      the unique name of the variable.
+    """
+    # Get the distributed variable if it exists.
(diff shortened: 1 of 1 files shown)


larq/optimizers.py
```diff
@@ -46,6 +46,24 @@
 __all__ = ["Bop", "CaseOptimizer"]
 
 
+def _var_key(var):
+    """Key for representing a primary variable, for looking up slots.
```
timdebruin, line 50: might be nice to have a link to the source, in case that changes

timdebruin (approved):
Nice!

larq/optimizers.py
lgeiger: Good idea, I added a link. This function hasn't changed since TF 1.14, so there is a good chance that we can keep the implementation as is.

leonoverweel:
Wow, tricky edge case! Glad to see `CaseOptimizer` is still being used :)

PR #811, opened 2022-11-04
Drop support for Python 3.6

lgeiger:
Python 3.6 is [EOL since 23 Dec 2021](https://endoflife.date/python). I think it is save to drop support for it in larq at this point.


Tombana (approved):

lgeiger:
I don't know why pytype fails on this PR but passes on main 🤷

CNugteren (approved):

PR #831, opened 2023-06-16
Support TensorFlow 2.13

lgeiger:


larq/quantized_variable_test.py
```diff
@@ -73,7 +73,8 @@ def test_method_delegations(distribute_scope):
         if version.parse(tf.__version__) > version.parse("1.14"):
             assert x.synchronization == x.latent_variable.synchronization
         assert x.aggregation == x.latent_variable.aggregation
-        assert evaluate(x.initialized_value()) == 7
+        if version.parse(tf.__version__) < version.parse("2.13.rc0"):
+            assert evaluate(x.initialized_value()) == 7
```
lgeiger, line 77: This internal function has long been deprecated and now throws a `NonImplemented` error.

larq/quantizers.py
```diff
@@ -718,9 +732,11 @@ def get(identifier):
     if identifier is None:
         return None
     if isinstance(identifier, dict):
-        return deserialize(identifier)
+        use_legacy_format = "module" not in identifier
+        return deserialize(identifier, use_legacy_format=use_legacy_format)
     if isinstance(identifier, str):
-        return deserialize(str(identifier))
+        config = {"class_name": str(identifier), "config": {}}
+        return get(config)
```
lgeiger, line 739: This applies the changes following [this commit](https://github.com/keras-team/keras/commit/7eb8ef28155f20cbd78e1e24b6da5cfb656d4164#diff-e69484e2f50471a5661ca897bd7ae336ea254a351904437d6ecdc6ab67ccd659R201-R207).

.github/workflows/unittest.yml
```diff
@@ -34,6 +34,10 @@ jobs:
             python-version: "3.8"
           - tf-version: 2.12.0
             python-version: "3.11"
+          - tf-version: 2.13.0
+            python-version: "3.8"
+          - tf-version: 2.13.0
+            python-version: "3.11"
```
CNugteren, line 40: Just checking, is this on purpose that we want to test it for two Python versions?

larq/quantized_variable_test.py
```diff
@@ -73,7 +73,8 @@ def test_method_delegations(distribute_scope):
         if version.parse(tf.__version__) > version.parse("1.14"):
             assert x.synchronization == x.latent_variable.synchronization
         assert x.aggregation == x.latent_variable.aggregation
-        assert evaluate(x.initialized_value()) == 7
+        if version.parse(tf.__version__) < version.parse("2.13.rc0"):
```
CNugteren, line 76: Should we just go with `2.13.0` here (and elsewhere in the code) instead of `rc0`?

.github/workflows/unittest.yml
lgeiger: This is mainly to match what we've been doing for other TF versions where we test against the oldest and newest Python version. But since 2.13 doesn't support Python 3.7 I've added this explicitly here matching what we do for TF 2.13.

I'm also happy to drop Python 3.7 since it's EOL since last week, but we probably should do that in a separate PR.

larq/quantized_variable_test.py
lgeiger: I don't really mind. I used `rc0` to make sure that this also works with the release candidates. But I'm happy to change it to `2.13.0` if you feel strongly about it.

CNugteren: Let's remove the `rc0`, forcing people to upgrade. I think no-one should be using the release candidate anymore, right? Furthermore, elsewhere in the code you do have `version.parse("2.13")` and `version.parse("1.15")`. Or would the solution be to also do `version.parse("2.13")` here and don't include `.0` to support the release candidates and keep the code clean?

lgeiger: > Or would the solution be to also do `version.parse("2.13")` here and don't include `.0` to support the release candidates and keep the code clean?

That won't be the case since `version.parse("2.13.0rc0") < version.parse("2.13") == True`

lgeiger: Done in 0707ae21b87fc83cbe04debf25fdd72b328de04c

CNugteren (approved):

PR #838, opened 2023-07-11
Add `QuantizedVariableSpec` to `QuantizedVariable`

lgeiger:
This adds a trace type to our quantized variable according to https://github.com/keras-team/keras/commit/ed99e34f279a2d2d6a44af87ee64f8fc98c7e8b9

This fixes an issue I am seeing on single GPU with optimizers that use `jit_compile=True`.

### DIFF
--- a/larq/quantized_variable.py
+++ b/larq/quantized_variable.py
@@ -2,6 +2,7 @@
 from typing import Optional
 
 import tensorflow as tf
+from packaging import version
 from tensorflow.python.distribute.values import DistributedVariable
 from tensorflow.python.framework import ops
 from tensorflow.python.ops import resource_variable_ops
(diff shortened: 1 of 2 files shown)


larq/quantized_variable.py
```diff
@@ -330,6 +366,11 @@ def shape(self):
     def get_shape(self):
         return self.latent_variable.get_shape()
 
+    def __tf_tracing_type__(self, context):
+        if _SUPPORTS_TRACE_TYPE:
+            return QuantizedVariableSpec(self)
+        return NotImplemented
```
DzedCPT, line 372: It's safe to return `NotImplemented` here? What does the caller do with the result?

DzedCPT:
> This adds a trace type to our quantized variable according to [keras-team/keras@ed99e34](https://github.com/keras-team/keras/commit/ed99e34f279a2d2d6a44af87ee64f8fc98c7e8b9)

Can we steal the test from that PR as well?

DzedCPT (approved):
Not super familiar with this code so take my approval with a pinch of salt, but it looks reasonable to me.

lgeiger:
> Can we steal the test from that PR as well?

Checking the dtypes doesn't really apply to us here. But I was able to make a minimal reproduction that would crash the tf function tracing without this fix.  I added it as a test in 983decb885d7f81247167f65de6da86461a84a0c

larq/quantized_variable.py
lgeiger: I think so, yes. If the TensorFlow version doesn't support trace type I am pretty sure this function will never get called so it's fine to return `NotImplemented`.

PR #916, opened 2026-05-19
Modernize repo: use ruff, drop Python 3.7, drop TF 1.x, etc.

CNugteren:
Modernize the larq repo:

  - Replace **black**, **flake8**, and **isort** with **ruff** for linting and formatting. Add `pyproject.toml` with ruff configuration. Drops the separate `lint` extras_require section.
  - Drop Python **3.7/3.8/3.9** support, require `>=3.10`. Remove `importlib-metadata` fallback in `larq/__init__.py`. Update TF minimum to `>=2.8.4`.
  - Update test dependencies: `pytest <8.4`, `pytest-cov <5.1`, `pytest-xdist <3.7`, `pytest-mock <3.15`. Bump `pytype` to `2024.10.11`. Bump `numpy` upper bound to `<3.0`.
  - Bump CI actions to v6 (`actions/checkout`, `actions/setup-python`). Switch from `toolmantim/release-drafter@v6` to `release-drafter/release-drafter@v7`. Use `pypa/gh-action-pypi-publish@release/v1`. Set dependabot schedule to 
  monthly.
  - Drop TF 1.x and 2.0–2.7 rows from the unittest matrix. Keep TF 2.8.4 through 2.17.0.
  - Fix 25 new ruff lint issues (B006, B007, B028, B904, B905, E721) across `context.py`, `layers_base.py`, `models.py`, `optimizers.py`, `quantized_variable.py`, `quantizers_test.py`, `testing_utils.py`, `utils.py`, and tests.
  - Pass `--config setup.cfg` to `pytype` explicitly (newer versions need it).

  Closes #893, #894, #900, #901, #902, #904 (these dependabot PRs are subsumed by the dependency bumps in this PR; #901 in particular is moot since flake8 is removed).


larq/layers_base.py
```diff
@@ -115,10 +115,11 @@ def __init__(self, *args, pad_values=0.0, **kwargs):
             raise ValueError("Causal padding with `pad_values != 0` is not supported.")
 
     def _get_spatial_padding_same(self, shape):
+        # Runs inside tf.function; autograph's zip wrapper does not accept `strict=`.
         return [
             _compute_padding(stride, dilation_rate, shape[i], filter_size)
             for i, (stride, dilation_rate, filter_size) in enumerate(
-                zip(self.strides, self.dilation_rate, self.kernel_size)
+                zip(self.strides, self.dilation_rate, self.kernel_size)  # noqa: B905
```
lgeiger: Let's maybe globally disable `B905` since this adds quite a lot of additional `noqa` statements

larq/utils.py
```diff
@@ -16,12 +16,12 @@ def memory_as_readable_str(num_bits: int) -> str:
...
     num_bytes = num_bits / 8
 
-    for i, suffix in enumerate(suffixes):
+    for i in range(len(suffixes)):
         rounded = num_bytes / (1024**i)
         if rounded < 1024:
             break
 
-    return f"{rounded:,.2f} {suffix}"
+    return f"{rounded:,.2f} {suffixes[i]}"
```
lgeiger: Let's revert these changes. What's wrong about enumerate

CNugteren: The issue here is that `suffix` isn't used in the loop and therefore the linter complains. But we do use it outside of the loop, which it doesn't understand. I've reverted the code and silenced the linter in this case.

larq/layers_base.py
CNugteren: Good idea, done.

setup.cfg
```diff
@@ -1,29 +1,11 @@
...
 inputs = larq
 output = .pytype
+exclude =
+    larq/snapshots
 # Keep going past errors to analyse as many files as possible.
 keep_going = True
 # Don't check use of imported modules, because we have no type-stubs for TF.
 strict_import = True
-python_version = 3.9
+python_version = 3.10
```
lgeiger: Can we also move all of this into pyproject.toml? Or might be good to switch to mypy or another typechecker since pytype seems to [not be developed anymore](https://github.com/google/pytype)

CNugteren: Done in https://github.com/larq/larq/pull/916/commits/b2b687071d0addb7a8229c476de2142e1acd6c49.

larq/models.py
```diff
@@ -341,7 +355,7 @@ def _generate_table_header(self, table_config: Mapping[str, Any]) -> Sequence[st
     def _generate_table_total(
         self, table_config: Mapping[str, Any]
     ) -> Sequence[Union[float, str]]:
-        row = ["Total", "", ""]
+        row: List[Union[float, str]] = ["Total", "", ""]
```
lgeiger: Since we now require Python 3.10 can we run [ruff's pyupgrade rule for 3.10](https://docs.astral.sh/ruff/rules/#pyupgrade-up) so we can use builtins for type definitions and replace Union with |

CNugteren: Done in https://github.com/larq/larq/pull/916/commits/7a81d1d924f65090243aba770ffa496e77105b74.

lgeiger (approved):

